PR #163, opened 2014-11-30
(feat) JUnit: Add support classes, including an `ApplicationTest` base test class.

hastebrot:
This PR adds some convenience classes for JUnit.

- It will be within the module `testfx-junit`.
- It will include an `ApplicationTest` class, that will serve as counterpart to JavaFX's `Application`. It launches the Toolkit per default and inherits methods from `FxRobot`.
- <s>It will also include helpful classes for JUnit `Category`s and `Rule`s.</s>

Issue #66, opened 2014-02-11
Allow custom stage to be set on GuiTest


aalmiray:
Currently `GuiTest` creates a `Stage` as a convenience for developers. But what if I already have a `Stage` instance setup? I'd love to be able to tell GuiTest to use that Stage (and hence it's root node) instead of what's done today. This may be related to https://github.com/SmartBear/TestFX/pull/39 asthis feature further enables reusability and composition.

Baba-sadou:
Hi,
I would like to  use TestFx for testing an application that has a preloader.
i launch this application with this instruction in my Main Class :
 LauncherImpl.launchApplication(MainApp.class, MainSplash.class, args);
i have looked for functions in FxToolKit class that can do the same thing, but i haven't found any.
could you help me please ?

hastebrot:
`LauncherImpl` does not support preloaders. You could implement support for preloaders youself, the functionality is in `launchApplicationWithArgs()` in `com.sun.javafx.application.LauncherImpl`. This methods calls `startToolkit()` which is problematic for testing, since we need to setup `Application`s multiple times, but starting the toolkit is only allowed once. This means we can't reuse this method for test setup, but we can copy parts of this method to launch preloaders and the application.

Baba-sadou:
ok, thinks

kayz1:
@hastebrot any news on launching applications with preloaders? Is there any example, anywhere? Thx

hastebrot:
@kayz1 A starting point would be here: https://github.com/javafxports/openjdk-jfx/blob/jfx-12/modules/javafx.graphics/src/main/java/com/sun/javafx/application/LauncherImpl.java#L135-L157

kayz1:
@hastebrot I managed to get it running with preloader, simply by setting 
`System.setProperty("javafx.preloader", ApplicationPreloader.class.getCanonicalName());`

Now I'am puzzled by cycles (testing, application). What is the correct approach for managing entire application, from start to shutdown.

Where do I start the app? In init - with launch(app.class)?
In @BeforeStart static method?
Can I stop the application for every test?

Any help or direction would be much appreciated.

PR #183, opened 2015-01-21
(feat) Allow instrumenting entire Applications.

0x4a616e:
We use TestFX for integration testing our JavaFX Application. Since we like to test the interoperability of all our UI components and the respective controllers, we need to test the application as a whole. Therefore, we refactored TestFX a little bit to be able to access the stage of custom Applications.

The biggest changes are:
* `FxToolkit` is no longer static, to create a FxToolkit with custom context
* `GuiTest` is split into `GuiTest` and `ApplicationTest` to test individual scenes and applications
* `ToolkitApplication` got a "delegate" Application, to which API calls are forwarded. This is `ToolkitDummyDelegate` for existing scene tests and can be any custom application that should be testes


hastebrot:
Hi, thanks for the PR!

I skimmed over your changes. I wonder if you uses cases, that need an introduction of a delegate `Application` and removal of static methods in `FxToolkit` could be archived with the present API.

- **Delegate application:** There is `FxToolkit.tookitContext().setApplicationClass()` that accepts a `Class<? extends Application>` for toolkit initialization. Or `FxToolkit.setupApplication()` and `FxToolkit.cleanupApplication()` for calling `init()`, `start()`, `stop()` on custom applications.

- **Removal of static methods:** There is `org.testfx.toolkit.ToolkitService` which is basically the non-static version of `FxToolkit`.

Is there a small use-case example that we could put into the `org.testx.cases.acceptance` package?

0x4a616e:
Thank you very much for pointing me into the right direction ;-) What you proposed is indeed much easier, so I changed the code accordingly.

There is also a small example in `org.loadui.testfx.CustomAppTest`. Unfortunately, I needed to exclude this from the remaining tests, as it tries to create a new app while the app used by the other tests is still running...

hastebrot:
I banished `GuiTest` into the subproject `testfx-legacy` and removed dependency to JUnit from `testfx-core`.

Work on this (i.e. ideas from `ApplicationTest`) should go into `testfx-junit` which is incubating in #163. Also JUnit `@Rule` fields should be explored.

0x4a616e:
:thumbsup: let me know if you want me to change anything!

0x4a616e:
Moved `ApplicationTest` and `FxTest` class into `testfx-legacy`

hastebrot:
Hi Jan, I think about changes to the introduction of `ApplicationTest` and `FxTest` in addition to `GuiTest`.

First of all `testfx-legacy` was introduced to ease the migration from `GuiTest` to the classes in `org.testfx.api`. After all, `GuiTest` was promoted as a general idea behind TestFX. It takes a lot of cognitive load away from testers, but it is limited to certain use-cases and bound to JUnit. That's why the new API was created.

Now there are a couple of reasons why I tend to remove your additions (we should find another way to allow your use-case):
- It moves the static methods `find()` and `findAll()` away from `GuiTest` to `FxTest`. There are no problems with TestFX's internal regression tests, because the former extends from the latter. But there could arise usability problems with IDEs, that might add static imports to `FxTest` instead of `GuiTest`, which adds coupling and is undocumented. Test developers with old tests are not affected, but we have to assume that `GuiTest` is used for new tests, since this way was promoted.
- It adds more classes in the main package `org.loadui.testfx` to choice from, which could cause confusion to beginners.
- It adds one more class to the inheritance hierarchy of the test class (i.e. `MyFancyTest` &raquo; `GuiTest` &raquo; `FxTest` &raquo; `FxRobot` &raquo; `FxRobotInterface`). Yes, `FxRobot` is designed to allow a choice between class inheritance and object aggregation. You can use inheritance to have short calls to robot methods (`clickOn()` et al.) and for compatibility reasons to use it in `GuiTest`. You can also use aggregation to be able to use robot methods in Spock `Specification`s or to use it in `@Rule`s for a non-intrusive setup as suggested by @aalmiray in #66.

I think we should have the classes in `org.testfx.api` to allow bare-bone testing for all general use-cases and keep `GuiTest` the only abstract test class in `testfx-legacy` for the use-cases it was supposed to.

Then there should be JUnit (and TestNG, and Spock, and Cucumber, ...) support classes the allow convenient testing with classes like `FxRobotTestBase`, `ApplicationTestBase` or `SceneTestBase`. And we should definately include `@Rule`s (I'm :100: behind this idea; we should promote the idea of non-intrusive setups).

0x4a616e:
Hi Benjamin, sounds reasonable! Is was looking through you're latest changes, and it appears to me that my use-case should already be possible by subclassing `TestCaseBase`. I'll try it out with our app, but it looks like `FxAssertBasicTest` is doing exactly what I was trying to achieve with my refactorings.

Using `@Rule`s sounds like a good plan, could be a good way to to setup the application for test :+1:

PR #213, opened 2015-04-21
added waitForRunLater method

caarlos0:
It's useful while testing stuff that is behind a Platform.runLater call,
  so it is not needed to add sleeps to tests.

Source:
http://stackoverflow.com/questions/22821129/wait-for-platform-runlater-in-a-unit-test

### DIFF
--- a/subprojects/testfx-junit/src/test/java/org/testfx/cases/acceptance/ApplicationRunLaterTest.java
+++ b/subprojects/testfx-junit/src/test/java/org/testfx/cases/acceptance/ApplicationRunLaterTest.java
@@ -1,0 +1,45 @@
+package org.testfx.cases.acceptance;
+
+import javafx.application.Platform;
+import javafx.scene.Scene;
+import javafx.scene.control.Button;
+import javafx.scene.layout.StackPane;
+import javafx.stage.Stage;
+import org.junit.Test;
+import org.testfx.api.FxToolkit;
+import org.testfx.framework.junit.ApplicationTest;
+
+import static org.testfx.api.FxAssert.verifyThat;
+import static org.testfx.matcher.base.NodeMatchers.hasText;
+
+public class ApplicationRunLaterTest extends ApplicationTest {
+    @Override
+    public void init() throws Exception {
+        FxToolkit.registerStage(() -> new Stage());
+    }
+
+    @Override
+    public void start(Stage stage) {
+        Button button = new Button("click me!");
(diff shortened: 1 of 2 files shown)


subprojects/testfx-junit/src/test/java/org/testfx/cases/acceptance/ApplicationRunLaterTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Test
+    public void should_contain_button() {
+        verifyThat(".button", hasText("click me!"));
+    }
+
+    @Test
+    public void should_click_on_button() throws Exception {
+        clickOn(".button");
+        waitForRunLater();
```
hastebrot: `clickOn()` should already wait for the JavaFX thread and events triggered by the mouse click event.

There is also `interact()` (member method of `ApplicationTest`) that allows custom interactions and waits for the JavaFX thread event queue afterwards.

caarlos0: didn't knew about that...

works for me, thanks!

hastebrot:
Hi Carlos,

thanks for the PR. Does `clickOn()` not wait for the user interactions? I think it's better to let the robot methods do the synchronization stuff automatically.

Here is a small list of methods that help with asynchronious processes (`#` denotes instance methods, `.` denotes static methods):

- `FxRobot#interact()`
- `FxToolkit.setupFixture()`, `FxToolkit.setupStage()`, ...
- `WaitForAsyncUtils.waitForFxEvents()` &mdash; Waits for the event queue of JavaFX Application Thread to be completed, as well as any new events triggered in it.
- `WaitForAsyncUtils.sleep()` &mdash; Sleeps the given duration.
- `WaitForAsyncUtils.waitForAsync()` &mdash; Runs the given `Runnable` or `Callable` on a new `Thread` and waits for a given duration to finish, otherwise times out with `TimeoutException`.
- `WaitForAsyncUtils.waitForAsyncFx()` &mdash; Runs the given `Runnable` or `Callable` on the JavaFX Application Thread at some unspecified time in the future and waits for it a given duration to finish, otherwise times out with `TimeoutException`.

subprojects/testfx-junit/src/test/java/org/testfx/cases/acceptance/ApplicationRunLaterTest.java
hastebrot: Ah, perfect. :smile_cat:

PR #234, opened 2015-11-23
(feature) Added ApplicationRule

johnzeringue:
Added `ApplicationRule`, a utility to allow rule-style integration with JUnit. The primary advantage of choosing this class over `ApplicationTest` is that using a rule doesn't preclude the user from integrating other test utilities, whether rules or base classes.


hastebrot:
Amazing! Thanks a lot.

hastebrot:
The nice thing is, that the `FxRobot` is directly available in the rule.

charlesrgould:
Nice!

PR #236, opened 2015-12-02
HeadlessPlatform doesn't seem to be necessary for headless tests

belu:
When migrating TestFX from 3.x to 4.x Alpha I noticed the following:

The package of HeadlessPlatform changed from JDK 1.8.0 Update 20 to 1.8.0 Update 40 ("headless" got removed from the package name). Therefore the constant has to be at least different when using JDK 1.8.0 Update 40 or above.

Interestingly for my setup headless testing even worked when removing the complete assignment code for the headless platform.


hillam:
@hastebrot ping!

brcolow:
I believe this was fixed by https://github.com/TestFX/TestFX/commit/606df553a578d99b4d05e83283bc94181d46ca6a . However, that does not address the concern of this PR which is that the code is unnecessary. Do I understand that correctly?

brcolow:
Closed: https://github.com/TestFX/TestFX/pull/346

hastebrot:
I wonder if this is true. Headless tests might run on some headless environments. I assume without `HeadlessPlatform` this will run the tests using graphical windows/dialogs when using a CI on Windows OS, i.e. parallel tests on the same machine will fail. Still have to prove this.

PR #237, opened 2015-12-04
(chore) Gradle: Upgrade to version 2.9 and cleanup scripts.

hastebrot:


brcolow:
I cherry-picked this commit and merged as part of a larger PR: https://github.com/TestFX/TestFX/pull/314
 Running a newer version of Checkstyle actually required a newer gradle wrapper version, hence the inclusion in that PR.

PR #251, opened 2016-03-07
(feat) ListViewMatchers: Add isEmpty().

brcolow:
Although one could get the same functionality as isEmpty() using:

`FxAssert.verifyThat(someListView, ListViewMatchers.hasItems(0));`

it better conveys what we are trying to assert/test by using:

`FxAssert.verifyThat(someListView, ListViewMatchers.isEmpty());`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,7 @@ atlassian*.xml
 
 # test screenshots.
 screenshot*.png
+
+# ctags
+tags
+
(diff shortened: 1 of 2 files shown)


hastebrot:
Thanks, for adding the matchers!

hastebrot:
An alternative method name could be `hasZeroItems()` to avoid any confusion.

PR #252, opened 2016-03-07
(feat) ListViewMatchers: Add hasPlaceholder(Node placeholder).

brcolow:
Ideal usage:

````java
FxAssert.verifyThat(lookup("#someListView"), ListViewMatchers.hasPlaceholder(new Label("List is empty!")));
````

### DIFF
--- a/subprojects/testfx-core/src/main/java/org/testfx/matcher/control/ListViewMatchers.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/matcher/control/ListViewMatchers.java
@@ -19,6 +19,7 @@ package org.testfx.matcher.control;
 import java.util.Objects;
 import javafx.scene.Node;
 import javafx.scene.control.Cell;
+import javafx.scene.control.Labeled;
 import javafx.scene.control.ListView;
 
 import org.hamcrest.Factory;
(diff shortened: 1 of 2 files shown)


brcolow:
A general question...

Does adding these types of methods make sense? Basically it is the difference between:

`assertThat(listView, ListViewMatchers.hasPlaceholder(new Label("Empty!")));`

and

`assertThat(listView.getPlaceholder(), LabeledMatchers.hasText("Empty!"));`

I am starting to think that the second way is better and would lead to less maintenance over time. Basically the strategy would be to have a XMatchers for every X that extends Node in JavaFX.

It would be really nice if there was a way to make this more "type-smart" in the sense of AssertJ, where you can do:

````java
List<String> myList = new ArrayList<>();
assertThat(myList). // <----- invoke completion here
````

and it shows you *type specific* assertions based on the type of the object inside the `assertThat` -  such as, for this example, [containsExactly](http://joel-costigliola.github.io/assertj/core/api/org/assertj/core/api/AbstractIterableAssert.html#containsExactly(T...)). This would take advantage of modern smart completion that IDEs have to aid the user in method discoverability. 

These two concepts that I have talked about are certainly not directly linked together, but I wanted to get my preliminary thoughts/musings out so that you could (hopefully) guide me in the best direction to take as we go forward. In summary, what do you think about the *idea* behind this PR and the [other one?](https://github.com/TestFX/TestFX/pull/251)? Are these legitimate additions? Should we only add methods to the XMatchers class (so in this case, since the placeholder property is a Node, it is already covered by NodeMatchers)?

Thanks!

hastebrot:
>Ideal usage:

~~~java
FxAssert.verifyThat(lookup("#someListView"), 
    ListViewMatchers.hasPlaceholder(new Label("List is empty!")));
~~~

For me the "idiomatic way" is:

~~~java
FxAssert.verifyThat("#someListView", 
    ListViewMatchers.hasPlaceholder("List is empty!"));
~~~

hastebrot:
>Or should we instead only add methods to the XMatchers class (so in this case, since the placeholder property is a Node, it is already covered by NodeMatchers)?

`NodeMatchers` contains some convienience methods like `hasText()`, because they might be used often. Maybe we could place `hasPlaceholder()` into `ControlMatchers`? I found `getPlaceholder()` in `ListView`, `TableView` and `TreeTableView`. `TreeView` does not have this method.

brcolow:
> For me the "idiomatic way" is:
> 
> FxAssert.verifyThat("#someListView", 
>     ListViewMatchers.hasPlaceholder("List is empty!"));

The problem is that this only works for placeholders that are Labeled. Consider, for example, a placeholder that is an ImageView.

PR #255, opened 2016-03-14
(feat) TableViewMatchers: Add containsRow().

brcolow:
This allows for testing that a TableView contains a given row at
a given row index. As an example, consider a Person class that has
three properties, name (StringProperty), age (IntegerProperty) and
birth-date (ObjectProperty<Instant>):

````java
TableView<Person> peopleTable = new TableView<>();
Person bob = new Person("Bob", 42, Instant.ofEpochSecond(139061613));
Person andy = new Person("Andy", 13, Instant.ofEpochSecond(1061260323));
Person jill = new Person("Jill", 5, Instant.ofEpochSecond(1320430636));
ObservableList<Person> people = FXCollections.observableArrayList(bob, andy, jill);

peopleTable.setItems(people);

// then:

assertThat(peopleTable, containsRow(0, "Bob", 42, Instant.parse("1974-05-29T12:13:33+00:00Z")));
assertThat(peopleTable, containsRow(1, "Andy" 13, Instant.parse("2003-08-19T02:32:03+00:00Z")));
assertThat(peopleTable, containsRow(2, "Jill", 5, Instant.parse("2011-11-04T18:17:16+00:00Z")));

// would all be true
````

### DIFF
--- a/subprojects/testfx-core/src/main/java/org/testfx/matcher/control/TableViewMatchers.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/matcher/control/TableViewMatchers.java
@@ -16,9 +16,14 @@
  */
 package org.testfx.matcher.control;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 import java.util.Objects;
+import javafx.beans.value.ObservableValue;
 import javafx.scene.Node;
 import javafx.scene.control.Cell;
+import javafx.scene.control.TableColumn;
 import javafx.scene.control.TableView;
 
 import org.hamcrest.Factory;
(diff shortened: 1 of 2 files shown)


hastebrot:
An alternative could be

~~~java
assertThat(from(peopleTable).lookupTableRow(0), 
    hasItems("Bob", 42, Instant.parse("1974-05-29T12:13:33+00:00Z")));
~~~

brcolow:
Hmmm..I played around trying to implement your suggestion but I am having serious troubles....essentially my problem is that I am not sure how to get the looked up row index from the hasItems method. To be honest, I am not sure I know how to implement your suggestion. Furthermore I am not sure it really is an improvement over the current approach. Maybe it would be an improvement to name the function: `nthRowContains(int n, Object...items)`? But to be frank I am in favor of the way it currently is - you let me know how you want to proceed.

hastebrot:
So let's merge this. We have tests and can refactor the method later, if this is needed.

My idea was, 

- to implement `lookupTableRow(rowIndex)` for `NodeQuery` which queries a `javafx.scene.control.TableRow`,
- to let `hasItems()` accept the `TableRow`, fetch the cells and compare them with the `Object...` array.

Another note: `hasItems(...)` could clash with the meaning of `TableView::getItems()`; both sound similar.

brcolow:
I think merging is the best way forward, maybe with a note on the commit itself or just your note above for history's sake :)

PR #257, opened 2016-03-21
(refactor) ApplicationService: Use Application instead of ApplicationFixture.

hastebrot:
- Use `Application` (instead of `ApplicationFixture`) in `ApplicationService` and `FxToolkit`. 
- Move `ApplicationFixture` and `ApplicationAdapter` from `testfx-core` to `testfx-junit`.

This changes `ApplicationAdapter` to extend `Application` instead of implementing `ApplicationFixture`. We need a `ApplicationFixture` interface in `testfx-junit` to let `ApplicationRule` extend `FxRobot` (single inheritance of Java).

~~~java
// before.
class ApplicationAdapter implements ApplicationFixture {}

// after.
class ApplicationAdapter extends Application {}
~~~

cc: @johnzeringue (this should not affect the behaviour of `ApplicationRule`).


hastebrot:
This also allows us to implement a complete `ApplicationServiceImplTest`.

PR #259, opened 2016-03-22
(feat) TextFlowMatchers: Add matchers for the text of a TextFlow.

brcolow:
Currently there are two ways to match the text contained in a TextFlow.
The first one is based simply off the contained text. The second is
based on the contained text *and* the color of that text. The expected
colors are encoded in the expected String like:

`<COLOR>some text</COLOR>`

where COLOR is one of the named colors in JavaFX. For a list of all such
colors, see:

https://docs.oracle.com/javafx/2/api/javafx/scene/doc-files/cssref.html#typecolor

We also extracted out a method for computing the distance between two
colors from PixelMatcherRgb as that functionality was also needed for
finding the color nearest to a named color.

---------------------------------------------------------

I am open to suggestions on a better way to test for colored text in a text flow as I just thought of the simplest thing I could. I could see the TextFlowMatchers being very useful for testing things that mimic
text editor functionality or also in cases where color-coded text helps the user understand the text (for example, using red for negative monetary balances and green for positive monetary balances) and provide better UX.

Let me know what you think!

### DIFF
--- a/subprojects/testfx-core/src/main/java/org/testfx/matcher/control/TextFlowMatchers.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/matcher/control/TextFlowMatchers.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright 2013-2014 SmartBear Software
+ * Copyright 2014-2015 The TestFX Contributors
+ *
+ * Licensed under the EUPL, Version 1.1 or - as soon they will be approved by the
+ * European Commission - subsequent versions of the EUPL (the "Licence"); You may
+ * not use this work except in compliance with the Licence.
+ *
+ * You may obtain a copy of the Licence at:
+ * http://ec.europa.eu/idabc/eupl
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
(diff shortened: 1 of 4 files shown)


hastebrot:
Is it possible to use the CSS color names of `javafx.scene.paint.Color` directly?

brcolow:
I think these a really good questions and valid feedback, I will look into them hopefully within the next 24 hours or so. Thanks Benjamin.

brcolow:
I updated the PR so that we now have two functions for testing the color of the text of a TextFlow:

`hasColoredText`:
Given the markup: `<COLOR>text</COLOR>`
and a TextFlow:
````java
Text text = new Text("text");
text.setFill(someColor);
TextFlow textFlow = new TextFlow(text);
````
Matches iff the named color that has the least distance from someColor is COLOR.

`hasExactlyColoredText`:
Given the markup: `<COLOR>text</COLOR>`
and a TextFlow:
````java
Text text = new Text("text");
text.setFill(someColor);
TextFlow textFlow = new TextFlow(text);
````

Matches iff someColor and COLOR are both named colors and they have exactly the same RGB values.

PR #267, opened 2016-05-05
Added moveTo(Node,offset)

Ortner:
Added a method to move to a point within a node with a given offset to the upper left corner of the node.

It's just a convenience method to prevent calls of moveTo(Node) and moveBy(Offset).

Other modifications:
- Added some Documentation (don't know whether it conforms to your JavaDoc guidelines, just dump it if not)
- Moved some methods as default implementation to the FXRobotInterface

By the way, why are the pointOfVisibleNode(...) methods private in FXRobot and not public API. I think they might be useful for users too.

### DIFF
--- a/subprojects/testfx-core/src/main/java/org/testfx/api/FxRobot.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/api/FxRobot.java
@@ -1113,47 +1113,6 @@ public class FxRobot implements FxRobotInterface {
         return this;
     }
 
-    @Override
-    @Unstable(reason = "is missing apidocs")
-    public FxRobot moveTo(double x,
-                          double y) {
-        return moveTo(point(new Point2D(x, y)));
-    }
-
-    @Override
-    @Unstable(reason = "is missing apidocs")
-    public FxRobot moveTo(Point2D point) {
-        return moveTo(point(point));
-    }
-
-    @Override
-    @Unstable(reason = "is missing apidocs")
-    public FxRobot moveTo(Bounds bounds) {
-        return moveTo(point(bounds));
-    }
-
-    @Override
-    @Unstable(reason = "is missing apidocs")
-    public FxRobot moveTo(Node node) {
-        return moveTo(point(node));
-    }
-    @Override
-    @Unstable(reason = "not tested yet")
-    public FxRobot moveTo(Node node, Point2D offset) {
-        return moveTo(point(node,Pos.TOP_LEFT).atOffset(offset));
-    }
-
-    @Override
-    @Unstable(reason = "is missing apidocs")
-    public FxRobot moveTo(Scene scene) {
-        return moveTo(point(scene));
-    }
-
-    @Override
-    @Unstable(reason = "is missing apidocs")
-    public FxRobot moveTo(Window window) {
-        return moveTo(point(window));
-    }
 
     @Override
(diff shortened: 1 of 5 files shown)


hastebrot:
Hey, sorry for not responding earlier.

>Added a method to move to a point within a node with a given offset to the upper left corner of the node.

There is `moveTo(offset(node, offsetX, offsetY)` which should do the same as `moveTo(Node node, Point2D offset)`. `offset()` was introduced in TestFX 3.x, and I guess the original author had the intention to minimize the need for additional methods with it. When we introduce `moveTo(node, offset)` we also need to introduce `clickOn(node, offset)`, `clickOn(string, offset)`, `clickOn(matcher, offset)`, ..., `doubleClickOn(node, offset)`, because one could also have the use-case to shorten the method calls.

>Added some Documentation (don't know whether it conforms to your JavaDoc guidelines, just dump it if not)

Many thanks. Looks good, will pull these in.

>Moved some methods as default implementation to the FXRobotInterface

What's the reasoning behind transforming the `FxRobot` methods into `FxRobotInterface` default methods?

>By the way, why are the pointOfVisibleNode(...) methods private in FXRobot and not public API. I think they might be useful for users too.

I'll consider making them public, or even pulling them into a service class.

Ortner:
Hey, no need to excuse, the answer was pretty fast in my opinion.

> There is moveTo(offset(node, offsetX, offsetY) which should do the same as moveTo(Node node, Point2D offset). offset() was introduced in TestFX 3.x, and I guess the original author had the intention to minimize the need for additional methods with it. When we introduce moveTo(node, offset) we also need to introduce clickOn(node, offset), clickOn(string, offset), clickOn(matcher, offset), ..., doubleClickOn(node, offset), because one could also have the use-case to shorten the method calls.

There is a small difference in those calls
- moveTo(Node) refers to the center of a Node - what is also my intention, when I call this method.
- moveTo(offset(node, offsetX, offsetY)) adds that offset to the center of the referred Node - but my intention, when using a offset within a node is usually to refer to the top left edge of that node.

So currently, the standard solution (I found) would be `moveTo(point(node).atPosition(Pos.TOP_LEFT).atOffset(offset))`. As I use this method quite frequently this expression is a bit unhandy. Is there a smarter solution right now? I couldn't find a nice one, but I'm pretty new to JavaFx.

I did also add the method point(Node, Position). It's already a bit shorter, but in my opinion still a bit too long (`moveTo(point(node,Pos.TOP_LEFT).atOffset(offset))`) for the standard use case. 

But you're right, the reference option would make sense in the other methods too. And the offset function seems to be the better place to implement it. I guess changing the standard offset method to refer to the top left edge is not an option. May be `offset(Node, Pos, offset)`? What would you suggest?


> What's the reasoning behind transforming the FxRobot methods into FxRobotInterface default methods?

Well, when a interface method does only use it's own public API without the need of internal state within the method itself, the method implementation should work for all implementations.
By makeing it a default method, it keeps the code in the implementing classes a bit more focused on the specific implementation details and eliminates the need of implementing common methods.

But if you do not use default methods for this, I'll put them back to the `FXRobot` class.

brcolow:
I like the idea of `offset(Node, Pos, offset)` personally. Maybe:

```java
public PointQuery point(Node node, Pos offsetReferencePos, Point2D offset);
```

For naming?

PR #268, opened 2016-05-05
Added interactFast(...) Method to FxRobot(Interface)

Ortner:
The interactFast method doesn't wait for the gui pulse which makes it
much faster. It is especially useful for timing relevant tests e.g.
animations and for reading parameters of the gui thread.

### DIFF
--- a/subprojects/testfx-core/src/main/java/org/testfx/api/FxRobotInterface.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/api/FxRobotInterface.java
@@ -187,8 +187,51 @@ public interface FxRobotInterface {
     // METHODS FOR INTERACTION AND INTERRUPTION.
     //---------------------------------------------------------------------------------------------
 
+    /**
+     * Calls a runnable on the FX application thread and waits for it and
+     * consecutive events to execute. So changes to the gui triggered by the
+     * runnable will be performed when returned from this method.
+     * 
+     * @param runnable
+     *            the runnable
+     * @return this robot
+     */
     public FxRobotInterface interact(Runnable runnable);
+
+    /**
+     * Calls a callable on the FX application thread and waits for it and
+     * consecutive events to execute. So changes to the gui triggered by the
+     * callable will be performed when returned from this method.
+     * 
+     * @param callable
+     *            the callable
+     * @return this robot
+     */
     public <T> FxRobotInterface interact(Callable<T> callable);
+    
+    /**
+     * Calls a runnable on the FX application thread and waits for it to
+     * execute. It does not wait for other events on the fx application thread.
+     * So changes to the gui triggered by the runnable may not be performed when
+     * returned from this method.
+     * 
+     * @param runnable
+     *            the runnable
+     * @return this robot
+     */
+    public FxRobotInterface interactNoWait(Runnable runnable);
+
+    /**
+     * Calls a callable on the FX application thread and waits for it to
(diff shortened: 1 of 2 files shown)


hastebrot:
Thanks for the PR.

Yes, we definitely need this method.

I'm not sure if `interactFast()` is the right name. I'd consider `interactNoWait()`, because `...NoWait()` is internally used for methods like this. However this could confuse the meaning, since the unrelated `waitFor()` is used in this method.

Ortner:
I guess with the Callable it's not a problem (as the result is returned), but in the combination with the Runnable as parameter, I thought that the first impression would be that we do not wait for the Runnable to execute. I assume no one would think of a interact method just using `Platform.runLater(...)`, so less waiting wouldn't make sense in this method. The javadoc explicitly mentions that it will be waited for the execution of the Runnable/Callable. So NoWait() should be OK.

I'm going to change the name.

Ortner:
By the way, any tests required for integration? If so, could you point me to the file where to implement it?

subprojects/testfx-core/src/main/java/org/testfx/api/FxRobotInterface.java
```diff
@@ -187,8 +187,45 @@ public PointQuery offset(String query,
     // METHODS FOR INTERACTION AND INTERRUPTION.
     //---------------------------------------------------------------------------------------------
 
+    /**
+	 * Calls a runnable on the FX application thread and waits for it and consecutive events to
```
brcolow: The JavaDoc comment leading stars look misaligned, they should be aligned like so:

```java
/**
 * 
 * Blah blah.
 *
 * @param blah
 * @return blah
 */
```

subprojects/testfx-core/src/main/java/org/testfx/api/FxRobotInterface.java
```diff
@@ -187,8 +187,45 @@ public PointQuery offset(String query,
...
+    /**
+	 * Calls a runnable on the FX application thread and waits for it to
+	 * execute. It does not wait for other events on the fx application thread.
+	 * So changes to the gui triggered by the runnable may not be performed when
+	 * returned from this method.
+	 * 
+	 * @param runnable the runnable
+     * @return this robot
+     */
+    public FxRobotInterface interactNoWait(Runnable runnable);
```
brcolow: Missing space between new methods.

brcolow:
I restarted the Travis CI build so that you can see the lint errors. I know its been a long time, but are you still interested in moving forward with this PR? Let me know, thanks very much @Ortner .

Ortner:
All done.

brcolow:
Thanks very much, could you squash the 4 commits into one with an updated commit message? I will then merge immediately.

Ortner:
I can try (but need to find a HowTo first :-).

brcolow:
```sh
git rebase -i HEAD~4
# The editor you have configured will popup
# Inside that editor, there will be 4 commits all proceeded by the word 'pick'
# Change the bottom 3 picks to `fixups` and the first pick to `reword`
# Write the commit message accordingly, and save and quit your editor (e.g. `:wq` in vim)
git push --force
```

Ortner:
Guess it worked. I left out the merge of the master branch, as I assume I should not merge those in the commit.

brcolow:
Merged. Thanks a lot for your patience.

Ortner:
Thanks for integrating!

Issue #271, opened 2016-05-16
Disable TestFX for CI


Ortner:
## Issue description
Skipping test in JUnit dynamically on the CI server does not work.
```Java
    @Before
    public final void setUp() {
		//skipping tests if not doing ui tests
		org.junit.Assume.assumeTrue(doUITest);
    }
```
This is the preferred way, as test are being mared as skipped in the test report. The test will hang up waiting for the Application to start in `ApplicationTest.internalAfter()`.

## Solution
Add a static field, that disables testsing (may be reading an environmentvariable additionally).
Modify setup routines to skip any UI stuff when the variable is disabled.

PR #272, opened 2016-05-16
Disable TestFX for CI

Ortner:
See #271 

Added static field, so all tests can be disabled by that field. The field is evaluated in the `internalBefore()` and `internalAfter()` methods.

Additionally all tests can be skipped by setting the environment variable `UI_TEST` to anything else than true (if not set all tests will be executed).


Ortner:
Some more work has to be done. The other classes should be disabled too. Generally a JUnit configuration should be used to set the disabled flag.

Ortner:
PR not yet mature. It requires other fixes first (#280 and another one). I will propose again, when the other issues are solved.

brcolow:
@Ortner Now that #280 has been merged, would you like to re-open this PR?

Issue #276, opened 2016-05-17
Failing test in org.loadui.testfx.TextInputControlsTest


Ortner:
## Issue

The following tests are failing:

```
org.loadui.testfx.TextInputControlsTest > shouldClearText FAILED
    java.lang.AssertionError at TextInputControlsTest.java:47
        Caused by: java.lang.AssertionError at TextInputControlsTest.java:47
```

As they do not fail on the Build-Server or are skipped (at least I couldn't find anything in the protocols). I assume the issue depends on my specific platform:
- Operating system: Apple OSX ElCapitan (10.11.4)
- Java: Oracle JDK8 (build 1.8.0_91-b14)
## Behaviour of test

``` Java
    @Test
    public void shouldClearText() {
        clickOn(TEXT_FIELD).write("Some text");
        verifyThat(TEXT_FIELD, hasText("Some text"));

        push(TAB); // To change focus from the TextField.

        clearTextIn(TEXT_FIELD);
        verifyThat(TEXT_FIELD, hasText(""));
    }
```

The text is correctly typed and verified in the first two lines. Within the rest of the test the last character is removed. And the test fails at the last verify.
## Analysis

The method clearTextIn() in in TextInputControls uses the following code to clear the TextField:

``` Java
fx.push(CONTROL, A).push(BACK_SPACE);
```

The keyboard shortcut to select all text on a Mac is 'command+A' actually.
## Proposed Solution

Either implement a platform specific behaviour or use the platform independent solution provided by `public static void clearTextIn(TextInputControl textInputControl)`

JordanMartinez:
@Ortner I've just started using TestFX, so I'm not familiar with its inner workings. However, shouldn't the system dependent `fx.push(CONTROL, A)` be the system independent `fx.push(SHORTCUT, A)`?

Is there some code code in the inner workings that screws this up or something?

Ortner:
@JordanMartinez I actually didn't new about the shortcut keycode. But as there is no need in using input events to clear the TextField, in my opinion the method used in #277 is the better solution.

Do you see any reason for using KeyEvents?

JordanMartinez:
> Do you see any reason for using KeyEvents?

To some extent, yes and no. 
Yes: because it would realistically simulate what would occur if one interacted with the `TextInputControl`. 
No: because what's the point of spending unneeded time in event handlers since we already know that it will only delete the text?

[...]

PR #277, opened 2016-05-17
Proposed fix for 276: Failing test in org.loadui.testfx.TextInputControlsTest

Ortner:
## Issue

See #276 
## Implemented Solution

Using clearTextInputControl(testInputControl) instead of ctrl+A, del as this doesn't work on a mac.
## Comments

Now all test are running (even on a Mac :-) ).

I accidentially added some gradle files in one of the commits. I reverted one of them, but I would also favour the insertion of the eclipse plugin. But it is not really in the focus of this issue...


Ortner:
This commit seems to brak the headless build. Haven't tried the headless option on may system at the moment. I will try to check this.

Ortner:
Updated to current master.
Headless build seems to work now.

brcolow:
What are your thoughts about detecting the OS, and using either Ctrl-A or Command-A instead of calling the clearTextIn method directly?

Ortner:
I think that should work as well. Is it also implemented in the legacy version of TestFX? 

We had the discussion in #276. I personally prefer clearing the text directly.

brcolow:
The change to use `Shortcut`-A was [merged](https://github.com/TestFX/TestFX/pull/333) - so this PR now only adds the documentation - once it is rebased I will merge.

brcolow:
Done in: https://github.com/TestFX/TestFX/pull/335

Issue #281, opened 2016-05-19
Vulnerability in WaitForAsyncUtils


Ortner:
## Issue

There are several vulnerabilities in the WaitForAsyncUtils. Public methods require Exception handling, where it is not obvious.

The following code snippet serves as example:

``` Java
try{
    Future<?> f=WaitForAsyncUtils.async(() -> {
        WaitForAsyncUtils.async(() -> {throw new SecurityException("Exception");});
    });
}catch(Exception e){e.printStackTrace();}
```

This code does not throw any exceptions.  
If you do not call the `get()` method of the Future returned by `async()`, exceptions during execution will be masked in the `callCallableAndSetFuture` method. But when using the `async()` call I actually do not want to wait for the result. 

I'm not completely done yet, as the Exception handling is not that easy to follow and different types of execution models with different kinds of exception handling are used (Runnables, callables with/without Executors...). But I think this is at least true for all the `async..(...)` methods. The `waitForAsync...(...)`Methods do check the exceptions later.

In my opinion this is dangerous, especially in the public API.
## Current test case

The current test case is also invalid, the Exception is thrown by another method (see comment below).

``` Java
@Test(timeout=1000)
    public void async_callable_with_exception() throws Exception {
        // given:
        Callable<Void> callable = () -> {
            throw new UnsupportedOperationException();
        };
        Future<Void> future = WaitForAsyncUtils.async(callable);

        // expect:
        thrown.expectCause(instanceOf(UnsupportedOperationException.class)); // <-- Validates after test
        WaitForAsyncUtils.waitFor(50, MILLISECONDS, future);  //<-- Exception is thrown here
    }
```
## Possible relations
- GlassRobot uses this methods quite frequently without checking the return value (move mouse, release mouse, key press...)
- As exceptions may be masked, unpredicted behavior (e.g. concurrent modification exceptions may be masked), so this issue might be related to  #279
## Proposal

I'm not completely done yet, but the simplest solution would be to remove the try catch block in `callCallableAndSetFuture`, if called from one of the async methods (extra method for async calls).

PR #282, opened 2016-05-21
281 vulnerabilityin wait for async utils

Ortner:
## Issue

See #281 
## Implemented Solution

Added a `ASyncFXCallable<X>` - this class does almost all the async stuff required. Exceptions are still fetched in the ExecutorService (or fired on the FX-Thread). Which leaves the JUnit thread running without exception.

There are two ways, to receive those Exceptions, either by calling `checkException()` or by calling any of the `async` methods again (if the field `autoCheckException` is left true).
Only not uncaught Exceptions will be returned by this method. So if Exceptions are handled in the Application, between subsequent calls of the `async` methods, no exceptions will be thrown.

This feature is turned off for the `waitFor` part, as exceptions will be checked internally.


brcolow:
I have taken care of the above in https://github.com/TestFX/TestFX/pull/351 so it is no longer necessary for you to do anything. Thanks very much for the PR. I am still working through the changes to wrap my head around them.

brcolow:
Merged as part of https://github.com/TestFX/TestFX/pull/351 - thanks!

PR #285, opened 2016-05-25
Add scroll left/right to ScrollRobot

### LINKED ISSUE #283: ScrollRobot should support horizontal directions
`org.testfx.robot.ScrollRobot` only supports scrolling up and down.

aalmiray:
Add support for horizontal scrolling to ScrollRobot and FxRobot.

Fixes #283

### DIFF
--- a/subprojects/testfx-core/src/main/java/org/testfx/robot/ScrollRobot.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/robot/ScrollRobot.java
@@ -24,14 +24,14 @@ public interface ScrollRobot {
     void scroll(int amount);
 
     void scroll(int positiveAmount,
-                       VerticalDirection direction);
+                VerticalDirection direction);
 
     void scrollUp(int positiveAmount);
 
     void scrollDown(int positiveAmount);
 
     void scroll(int positiveAmount,
-                       HorizontalDirection direction);
+                HorizontalDirection direction);
 
     void scrollLeft(int positiveAmount);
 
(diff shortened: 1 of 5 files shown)


subprojects/testfx-core/src/main/java/org/testfx/robot/ScrollRobot.java
```diff
@@ -16,17 +16,24 @@
...
                        VerticalDirection direction);
 
-    public void scrollUp(int positiveAmount);
+    void scrollUp(int positiveAmount);
 
-    public void scrollDown(int positiveAmount);
+    void scrollDown(int positiveAmount);
 
+    void scroll(int positiveAmount,
+                       HorizontalDirection direction);
```
brcolow: Weird indentation on the second argument here.

aalmiray: Fixed with https://github.com/TestFX/TestFX/pull/285/commits/9fb9674fb4b8250f658107d4b37dc4f8ee751b4a

brcolow:
Merged. Thanks a lot!

aalmiray:
lovely! Eagerly awaiting the next release 😄

JordanMartinez:
@aalmiray I am adding documentation in #325 and I came across your work here. When I looked at the the actual code that gets called in `GlassAdapterRobot`, I don't understand how the system is able to distinguish a scroll up from a scroll left or a scroll down from a scroll right. Care to explain?

aalmiray:
The original implementation of the scroll methods simply incremented the current position by 1. This would scroll to the right if the orientation is horizontal, down if vertical. The new ScrollRobot class adds left/up support by  the position by -1.

JordanMartinez:
Ah... so orientation is what determines that... I didn't understand how passing a value of 1 could scroll both down and right.

PR #288, opened 2016-05-27
(fix) Add check to cleanup for active FX Application Thread.

### LINKED ISSUE #247: Platform.exit()
When I use ```Platform.exit()``` at points where I want to force a direct exit of the application, my unit tests with TestFX seem to hang. I'm guessing they're not shutting down properly.

Is it a bad idea using ```Platform.exit()```, or can this be fixed within TestFX somehow?

### LINKED ISSUE #287: (bug) Cleanup hangs when FX Application Thread got killed
## Issue
The Cleanup is called in the `after` method of JUnit tests. If the FX Application Thread is beeing killed during the test, the call to the clean up method `cleanupApplication()` in `FXToolkit` waiting for the cleanup to be performed on the FX Application Thread.

## Solution
Detect, if the Application thread is running, before doing the cleanup.

Ortner:
## Issue
see #287 and #247 

## Solution
Added method `isFXApplicationThreadRunning()` in `FxToolkit` to detect whether FX Application Thread is running. Clean up code is only pushed  on the FX-Thread if it is running.

fixes #287
fixes #247

### DIFF
--- a/subprojects/testfx-core/src/main/java/org/testfx/api/FxToolkit.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/api/FxToolkit.java
@@ -180,10 +180,16 @@ public final class FxToolkit {
         );
     }
 
+    /**
+     * Performs the clean up of the application. This is done by calling {@link ToolkitService#cleanupApplication(Application)}
+     * (which usually calls the {@code stop} method of the application.
+     * @param application the application to clean up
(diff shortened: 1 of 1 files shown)


Ortner:
Modified the solution a bit:
Direct call to `isFXApplicationThreadRunning()` from `cleanupApplication`, `service.cleanupApplication(application)` doesn't get called at all.

PR #303, opened 2016-08-29
Allow for matching a table row without specifying index.

brcolow:
Rename the old containsRow method to containsRowAtIndex and add a new
method, containsRow, that matches if the given TableView has the row
at any index.


brcolow:
Ping @hastebrot

PR #308, opened 2016-10-19
(docs) fix pom formatting

feldoh:
Slight formatting fix. The extra indent made me initially mis-read it; so I thought I'd fix it.


brcolow:
Merged. Thanks!

PR #312, opened 2016-11-29
Add building on macOS to travis.

brcolow:
First attempt at getting Travis to also build on OSX, and add gitter webhooks.


brcolow:
I am doing something incorrectly with the `allow_failures` bit, because Travis is not picking up the fact that the osx build with `OSX_HEADLESS=false` should be allowed to fail.

brcolow:
I don't understand why the allowed failures are not getting picked up.

PR #319, opened 2016-12-02
Release all pressed keys after test finishes to prevent hanging

### LINKED ISSUE #295: Infinite KeyPressed event loop
Hi,

I'm launching a Test with JUnit on a Linux Debian machine inside a Eclipse Mars IDE.

``` Java
@Test
public void writeInTextFieldAndPressEnter()
{
    clickOn("#myNode").write("something").press(KeyCode.Enter);
}
```

The test run successfully but when it stops, the Enter Key continue to raise KeyPressed Event. That's to say, if I open a new window like Gedit, a lot of new line characters are inserted inside it.

When I finally stop the Eclipse IDE, the KeyPressed Event stops being raised.

Do you have any clue of what happen ?

JordanMartinez:
Should fix #295.

### DIFF
--- a/subprojects/testfx-junit/src/test/java/org/testfx/framework/junit/KeyAndButtonReleaseTest.java
+++ b/subprojects/testfx-junit/src/test/java/org/testfx/framework/junit/KeyAndButtonReleaseTest.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2013-2014 SmartBear Software
+ * Copyright 2014-2015 The TestFX Contributors
+ *
+ * Licensed under the EUPL, Version 1.1 or - as soon they will be approved by the
+ * European Commission - subsequent versions of the EUPL (the "Licence"); You may
+ * not use this work except in compliance with the Licence.
+ *
+ * You may obtain a copy of the Licence at:
+ * http://ec.europa.eu/idabc/eupl
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the Licence is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the Licence for the
+ * specific language governing permissions and limitations under the Licence.
+ */
+package org.testfx.framework.junit;
+
+import javafx.scene.input.KeyCode;
+import javafx.scene.input.MouseButton;
+import javafx.stage.Stage;
+
+import org.junit.FixMethodOrder;
+import org.junit.Test;
+import org.junit.runners.MethodSorters;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+
+/**
+ * Tests to see if pressed keys/buttons are released after a test is run.
+ *
+ * For keys, this prevents hanging issues (if Shortcut key is pressed) or deletion issues (if backspace/delete key
+ * is pressed) or insertion issues (if anything else is pressed). The same for mouse buttons
+ */
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
+public class KeyAndButtonReleaseTest extends ApplicationTest {
+
+    @Override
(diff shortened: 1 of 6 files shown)


brcolow:
Hey, thanks for this! Would you be so kind as to add a unit test that shows that, even when not explicitly called, the keys are freed? I imagine you could do this by creating a test file `testfx-unit/src/test/java/org/testfx/framework/junit/KeyReleaseTest.java` that is like so:

```java
@FixMethodOrder(MethodSorters.NAME_ASCENDING)
public class KeyReleaseTest {
    @Test
    public void A_test_that_forgets_to_release_keys() {
       // press some modifier keys (like shift, ctrl, or capslock)
   }
   @Test
   public void B_next_test_keys_are_not_pressed() {
       // Make sure the keys from the above test are not in the PRESSED state
    }
}
```

If you don't have time to do this I will try and do it in the next couple days. Let me know if you need more clarification, or if you have a better way to test for what this PR does. I think adding pseudo-regression tests like this will help to make sure we don't break things going forward!

JordanMartinez:
How would I implement the second test?

brcolow:
You mean how to check that they are no longer pressed? That's a good question, actually...

Maybe something like this?

```java

import org.junit.FixMethodOrder;
import org.junit.Test;

import static org.hamcrest.CoreMatchers.is;

@FixMethodOrder(MethodSorters.NAME_ASCENDING)
public class KeyReleaseTest  extends TestCaseBase {
    private Stage primaryStage;
    private EventHandler<? super KeyEvent> keyReleasedHandler;
    private final CountDownLatch keyReleasedLatch = new CountDownLatch(1);
    
    @BeforeClass
    public static void setupSpec() throws Exception {
        FxToolkit.setupStage(stage -> {
            primaryStage = stage;
            Scene scene = new Scene(new AnchorPane(), 300, 300);
            stage.setScene(scene);
            stage.show();
        });
        keyReleasedHandler = event -> keyReleasedLatch.countDown();
    }
    
    @Test
    public void A_test_that_forgets_to_release_keys() {
       // press some modifier keys (like shift, ctrl, or capslock)
   }
   @Test
   public void B_next_test_keys_are_not_pressed() {
       // For however many keys we pressed that should have been released
       // the keyReleasedLatch should be initialized to that size (I did
       // 1 above, this should be replaced with real number)
       assertThat(keyReleasedLatch.getCount() == 0, is(true));

    }
}
```

JordanMartinez:
Sorry but I found your example to be confusing because I don't see how the event handler gets called. Also, I'm not sure what `TestCaseBase` is...

I can't do any more tonight, but perhaps what I did come up with will work. If not, let me know.

brcolow:
Sorry, I forgot to attach the event handler to the scene in the above test case.

JordanMartinez:
@brcolow How's that? 
For some reason, my IDE wasn't able to initially detect hamcrest's `assertThat` method. I had to figure out the correct import by looking at another test that used it.

subprojects/testfx-junit/src/test/java/org/testfx/framework/junit/KeyAndButtonReleaseTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Override
+    public void start(Stage stage) throws Exception {
+        stage.show();
+    }
+
+    @Test
+    public void b_When_a_test_forgets_to_release_keys() {
+        press(KeyCode.CONTROL, KeyCode.SHIFT, KeyCode.ALT);
+    }
```
brcolow: Missing space between these two methods.

JordanMartinez: Fixed

brcolow:
Just waiting for CI to pass, and will merge!

JordanMartinez:
Excellent.

PR #320, opened 2016-12-02
Convert Shortcut to os-specific shortcut key

### LINKED ISSUE #290: KeyCode.SHORTCUT does not convert into System-dependent shortcut KeyCode
By principle, `ApplicationTest#press(SHORTCUT)` should run a `Node`'s `CONTROL` handler for Windows/Linux and the `COMMAND` for Mac

JordanMartinez:
Should fix #290, but I'm not exactly sure how it should be tested...

### DIFF
--- a/subprojects/testfx-junit/src/test/java/org/testfx/framework/junit/ShortcutKeyTest.java
+++ b/subprojects/testfx-junit/src/test/java/org/testfx/framework/junit/ShortcutKeyTest.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2013-2014 SmartBear Software
+ * Copyright 2014-2015 The TestFX Contributors
+ *
+ * Licensed under the EUPL, Version 1.1 or - as soon they will be approved by the
+ * European Commission - subsequent versions of the EUPL (the "Licence"); You may
+ * not use this work except in compliance with the Licence.
+ *
+ * You may obtain a copy of the Licence at:
+ * http://ec.europa.eu/idabc/eupl
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the Licence is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the Licence for the
+ * specific language governing permissions and limitations under the Licence.
+ */
+package org.testfx.framework.junit;
+
+import java.util.Locale;
+
+import javafx.scene.Scene;
+import javafx.scene.control.TextField;
+import javafx.scene.input.KeyCode;
+import javafx.stage.Stage;
+
+import org.junit.Test;
+
+import static javafx.scene.input.KeyCode.COMMAND;
+import static javafx.scene.input.KeyCode.CONTROL;
+import static javafx.scene.input.KeyCode.SHORTCUT;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+
+/**
+ * Tests whether pressing {@code KeyCode.SHORTCUT} will convert into the OS-specific KeyCode (i.e.
+ * {@code KeyCode.COMMAND} for Macs and {@code KeyCode.CONTROL} for everything else).
+ */
+public class ShortcutKeyTest extends ApplicationTest {
+
+    private final KeyCode osSpecificShortcutKey;
+
+    {
+        String osName = System.getProperty("os.name").toLowerCase(Locale.US);
+        osSpecificShortcutKey = osName.startsWith("mac") ? COMMAND : CONTROL;
+    }
+
+    private TextField field;
+    private final String pressedText = "pressed";
+    private final String releasedText = "released";
+
+    @Override
+    public void start(Stage stage) throws Exception {
+        field = new TextField();
+        field.setOnKeyPressed(e -> {
+            if (e.getCode().equals(osSpecificShortcutKey)) {
+                field.setText(pressedText);
+            }
+            e.consume();
+        });
+        field.setOnKeyReleased(e -> {
+            if (e.getCode().equals(osSpecificShortcutKey)) {
+                field.setText(releasedText);
+            }
+            e.consume();
+        });
+        stage.setScene(new Scene(field, 400, 400));
+        stage.show();
+        field.requestFocus();
+    }
+
+    @Test
+    public void shortcut_keyCode_converts_to_OS_specific_keyCode_when_pressed() {
+        // when:
+        press(SHORTCUT);
+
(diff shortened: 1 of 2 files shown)


JordanMartinez:
Also, when I ran `gradle test`, this test failed on my system
````java
org.testfx.cases.acceptance.ApplicationStartTest > should_click_on_button FAILED
    java.lang.AssertionError at ApplicationStartTest.java:71
````

Any idea why?
````
Linux Mint 18

java version "1.8.0_111"
Java(TM) SE Runtime Environment (build 1.8.0_111-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.111-b14, mixed mode)
````

**Comment by brcolow** [comment]
Does it fail consistently or is it flaky? As for how to test this, I will think about it and try and come up with something.

**Comment by JordanMartinez** [comment]
It's failed all 3 times I've tried it.

**Comment by JordanMartinez** [comment]
@brcolow any ideas for how I should test this? I could use a modification of the example you provided in #319 to change a TextField's text if the correct shortcut key is pressed. But if I do that, should the test still go in RobotImplTest.java? Or in the same package as KeysAndButtonsReleasedTest.java?

**Comment by brcolow** [comment]
I don't feel strongly as to which file it belongs in...I guess I would say RobotImplTest but again, I don't feel strongly on that. I think what you said is a good idea/plan.

**Comment by JordanMartinez** [comment]
Personally, I think it should stay in its own package as it tests BaseRobotImpl, which is in a different package altogether. However, since I don't know this code base very well yet, it is easier to implement it the test using ApplicationTest.

**Comment by brcolow** [comment]
In this case, I don't think it's necessary to force the method order to be ascending, as it is not testing the sequence of the two tests, is it? In other words, can the tests run in any order (A -> B, B ->A) and the outcome be the same? If so, we should remove the explicit ordering as ideally unit tests should be able to run in any order.

**Comment by JordanMartinez** [comment]
True. I forced the method order because I was going to use `press(SHORTCUT)` in the second one before later changing it to the OS-specific one. Give me a sec.

**Comment by JordanMartinez** [comment]
The 2nd test fails on my side now.

File: subprojects/testfx-core/src/main/java/org/testfx/robot/impl/BaseRobotImpl.java [hunk 1]
```diff
@@ -42,6 +42,13 @@
     private static final String PROPERTY_TESTFX_ROBOT_AWT = "awt";
     private static final String PROPERTY_TESTFX_ROBOT_GLASS = "glass";
 
+    private static final KeyCode OS_SPECIFIC_SHORTCUT;
+
+    static {
+        String osName = System.getProperty("os.name");
```
brcolow: Can you make this: `String osName = System.getProperty("os.name").toLowerCase(Locale.US)` and then below use

`osName.startsWith("mac")`

subprojects/testfx-junit/src/test/java/org/testfx/framework/junit/ShortcutKeyTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+import static javafx.scene.input.KeyCode.COMMAND;
+import static javafx.scene.input.KeyCode.CONTROL;
+import static javafx.scene.input.KeyCode.SHORTCUT;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+
+/**
+ * Tests whether pressing KeyCode.SHORTCUT will convert into the OS-specific KeyCode (i.e. COMMAND for Macs
```
brcolow: `KeyCode.SHORTCUT` -> `{@code KeyCode.SHORTCUT}`

brcolow:
I see two lint failures regarding unused imports...but there are also test failures on your end?

JordanMartinez:
Yup:
````java
org.testfx.framework.junit.ShortcutKeyTest > shortcut_keyCode_converts_to_OS_specific_keyCode_when_released FAILED
    java.lang.AssertionError: 
    Expected: "keyReleased"
         but: was "keyPressed"
        at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
        at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:8)
        at org.testfx.framework.junit.ShortcutKeyTest.shortcut_keyCode_converts_to_OS_specific_keyCode_when_released(ShortcutKeyTest.java:83)
````

**Comment by brcolow** [comment]
What if you add a:

```java
@Before
public void clearTextField() {
    field.clear();
}
```

Also, instead of doing:

`field.addEventHandler(KeyEvent.KEY_PRESSED,` and `field.addEventHandler(KeyEvent.KEY_RELEASED`
it might be more readble to do:

`field.setOnKeyPressed(...)` and `field.setOnKeyReleased(..)`

JordanMartinez:
Ah... I see what's the issue... KeyboardRobotImpl holds the `keysPressed`, right? So when I call `press(osSpecificKey)`, it stores the specific key. When I call `release(SHORTCUT)`, it doesn't contain the shortcut keycode, so it does nothing.

brcolow:
Ahhh, yes, good catch! When release(SHORTCUT) is pressed, maybe we should check if keysPressed contains the OS-specific key (either command or control) instead?

JordanMartinez:
Yes, I think we should. Otherwise, someone might accidentally do what I just did.

JordanMartinez:
These check style tests.... Sheez!

Ok. It built successfully on my side.

JordanMartinez:
I also moved the conversion step from `BaseRobotImpl` to `KeyboardRobotImpl`.

brcolow:
Awesome. Looks good. I know the Checkstyle checks are a bit excessive, but having a thorough and enforced style guide is worth the trade-off, at least, I think so :).

Thanks for your patience and diligence! Merged :).

brcolow:
Some of the annoyance can be mitigated by (if you use IntelliJ IDEA IDE) using the [checkstyle-idea](https://github.com/jshiell/checkstyle-idea) plugin. It allows you to setup a profile of rules (which is done by selecting the checkstyle rules file `TestFX/gradle/checkstyle.xml`. Then you can run Checkstyle against the current file, or even the entire project.

JordanMartinez:
Gasp! Lifesaver!

Hmm... Should have read through that guide so my commit messages better adhere to that style.

JordanMartinez:
> Reminder to myself to add the above suggestion to https://github.com/TestFX/TestFX/wiki/How-to-Contribute

I added a small tip in the check style section. It's at least something now that you can modify/improve later.

brcolow:
Awesome, I re-worked it a bit. Thanks again.

PR #323, opened 2016-12-02
(fix) Center the stage in the middle of the primary screen.

### LINKED ISSUE #307: Default stage position
I've had a number of test failures in Jenkins where controls not visible because stage at arbitrary position. This causing controls be pushed off edge of the physical screen meaning FxRobot could not get the mouse over them. 
This was with using Xvfb.

I've since added stage.setX(0);stage.setY(0); throughout my tests. 

Id like to see this as default testfx behaviour so stage always positioned at (0,0) or other configurable position.

brcolow:
Fixes #307


brcolow:
Would love to get your opinion on this @hastebrot

mvsoder:
@brcolow Could you update your branch to master and resolve any conflicts. This should kick of a build for it.

brcolow:
It's been so long I don't remember anything about this. I did try and use the Github web UI merge. Feel free to disregard.

mvsoder:
If it's alright with you, I'll just duplicate your work in a separate PR.

PR #328, opened 2016-12-03
Add javadoc to util package

JordanMartinez:

PR #329, opened 2016-12-03
(cleanup) Use `getWidth()` and `getHeight()` rather than calculate it

JordanMartinez:
Address a point I made in #328


brcolow:
Makes sense. Good catch!

PR #330, opened 2016-12-04
(refactor) Use Java 8 Streams, not Guava

JordanMartinez:


JordanMartinez:
Note: There is still one section of guava code because I wasn't sure how to implement the method correctly:
````java
private static <T> Set<T> combine(T input,
                                  Collection<Function<T, Set<T>>> functions) {
    return FluentIterable.from(functions)
        .transformAndConcat((f) -> f.apply(input))
        .toSet();
}
````

**Comment by brcolow** [comment]
Hmm..I am not too familiar with the Guava libraries and FluentIterable in particular.

But maybe something like this?

```java
import java.util.Collection;
import java.util.Set;
import java.util.function.Function;
import java.util.stream.Collectors;

    private static <T> Set<T> combine(T input, Collection<Function<T, Set<T>>> functions) {
        return functions.stream().map(function ->  function.apply(input)).flatMap(Collection::stream)
                .collect(Collectors.toSet());
    }
```

JordanMartinez:
Your code is correct, but the issue is that `Collectors.toSet()` does not care about the returned order of items in the returned set. Thus, the test that checks this method (`assertThat(results, contains(label0, label1, ... button0, button1)`) fails because `contains` checks order as well as contents. This can be resolved if `containsInAnyOrder` was used instead, but is that desired? Is order important?

brcolow:
Hmmm...are the "leaf-nodes" of the scene-graph in a well defined order? If so, then I guess order does matter and we could add a sort call. If the "leaf-nodes" of the scene-graph are not in a well-defined order, then `containsInAnyOrder` is more appropriate.

brcolow:
What I mean is, are the circled nodes in a well-defined order?

sgtree

JordanMartinez:
Yes, they seem to be in a well-defined order. 

This is what the FXML looks like:
````xml
<StackPane maxHeight="-Infinity" maxWidth="-Infinity" minHeight="-Infinity" minWidth="-Infinity" prefHeight="400.0" prefWidth="200.0" xmlns="http://javafx.com/javafx/8" xmlns:fx="http://javafx.com/fxml/1">
   <children>
      <VBox prefHeight="200.0" prefWidth="100.0">
         <children>
            <HBox fx:id="labels" spacing="20.0">
               <children>
                  <Label fx:id="label0" text="0" />
                  <Label fx:id="label1" text="1" />
                  <Label fx:id="label2" text="2" />
               </children>
               <padding>
                  <Insets bottom="10.0" left="10.0" right="10.0" top="10.0" />
               </padding>
            </HBox>
            <HBox fx:id="buttons" spacing="10.0">
               <padding>
                  <Insets bottom="10.0" left="10.0" right="10.0" top="10.0" />
               </padding>
               <children>
                  <Button fx:id="button0" mnemonicParsing="false" text="0" />
                  <Button fx:id="button1" mnemonicParsing="false" text="1" />
                  <Button fx:id="button2" mnemonicParsing="false" text="2" />
               </children>
            </HBox>
````

brcolow:
Maybe http://stackoverflow.com/questions/27611896/is-there-a-collector-that-collects-to-an-order-preserving-set is relevant?

JordanMartinez:
Still failed... :-/

I'll have to look into it another time. It's getting late.

JordanMartinez:
Fixed it. It works if I use a `LinkedHashSet` instead of a `HashSet`

brcolow:
Thanks! Merged.

PR #331, opened 2016-12-05
Document FxRobotInterface

JordanMartinez:


JordanMartinez:
There's one issue I find with this: one has to lookup the javadoc of the actual method and then look at the method it calls to actually see what it does. Otherwise, one would have to update two places of documentation.

Should this javadoc-by-referral be the case? Or should the javadoc be transferred from the `[Action]Robot` interface to the `FxRobotInterface`?

brcolow:
I think the documentation should be on the highest point up the inheritance chain, and then the documentation will be automatically `@Inherited`, for cases where we are not talking about implementing an interface or abstract class, we can use `@see org.testfx.Class#method` to keep the docs in one place.

JordanMartinez:
> I think the documentation should be on the highest point up the inheritance chain, and then the documentation will be automatically @Inherited,

Right.

> For cases where we are not talking about implementing an interface or abstract class, we can use @see `org.testfx.Class#method` to keep the docs in one place.

I believe this answers my question, which I interpret as "Continue doing as you've done."

On a separate note, I realize the wording on my second commit could be misunderstood to mean "I removed the `@Unstable` annotation." I'll need to fix that when I have time to do so.

JordanMartinez:
This is still a WIP

JordanMartinez:
Still a WIP, though it'll be done probably within 2 days because I'll be busy tomorrow and the following day's morning.

JordanMartinez:
K. This should be done now. Look it over and let me know if I need to make any changes anywhere.

brcolow:
:> -- awesome. Thanks so much!

PR #338, opened 2016-12-16
Replace all uber jar deps

SimY4:
TestFx uses compile dependencies like mockito-all, hamcrest-all, etc. which doesn't play well with other project dependencies since they basically are one big dependency which can't be versioned partially. 
I've replaced all these dependencies onto a more preferable ones.

Additionally I scoped junit dependency as provided, so user would have to provide their own preferable junit version to work with TestFX.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -74,6 +74,10 @@ rootProject.with { project ->
 //-------------------------------------------------------------------------------------------------
 
 allprojects { project ->
+    configurations {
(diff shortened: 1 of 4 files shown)


brcolow:
Makes sense to me. Thanks for this. The failure was just a sporadic macOS failure. After restarting the build, everything was good 👍 . Merged.

PR #339, opened 2016-12-18
Doc toolkit package

JordanMartinez:
Assuming checks pass, this is done.


JordanMartinez:
Well, `PrimaryStageFuture` could be documented, but it's essentially just a `Future` that can run listeners after it is set. If people follow the javadoc trail of its subclasses, they'll figure that out. So, I'm not sure if that class really needs to be documented.

brcolow:
Looks good to me, Appveyor failure is just flakiness (obviously as this is only comments :)). Thanks a ton.

PR #341, opened 2016-12-19
Doc api package

JordanMartinez:
WIP


JordanMartinez:
The only object left to document in this PR is the `FxAssert` class. However, it is annotated with `@Unstable` because "method signatures need fine-tuning." I feel like that should be resolved before writing any documentation. Otherwise, the first draft of documentation may become outdated once these method signatures are refined.

brcolow:
Sounds good to me. Does that mean this is ready?

JordanMartinez:
yeah. It's ready.

JordanMartinez:
actually, let me fix that previous commit message...

JordanMartinez:
Ok. Now it's ready once tests pass.

brcolow:
Thanks!

PR #342, opened 2016-12-19
(refactor) Moved `ShortcutKeyTest` to core package.

JordanMartinez:
Now that I have a better understanding of how the core package works, I thought this test should be moved to its actual package since its an implementation detail. 
The other test I wrote, `KeyAndButtonReleaseTest`, seems like it should stay in its respective packages as those packages have classes used for testing. Thus, hang-prevention should already be a part of them.


brcolow:
Awesome. Makes sense. Thanks.

JordanMartinez:
@brcolow When I look at the commit history of this project (not this PR), this commit appears with an "x" next to it because all the tests in `ToolkitServiceImplTest` failed. Is that test flaky? Or did my code not clean itself up correctly?

brcolow:
It is only flakiness. This is verified by later commits having green checkmarks. Those failures are quite hard to reproduce reliably and have something to do with headless macOS.

PR #344, opened 2016-12-21
(doc) Add documentation to `matcher` package.

JordanMartinez:
Rather than having one commit per class, I decided to group it together into one commit.

This is done.


brcolow:
I like the single commit as it keeps the history very clean. Thanks a lot.

JordanMartinez:
Aye, I should have done that on all the other ones :-/ I take it we can't overwrite the master branch with an updated version that squashes those commits?

brcolow:
We technically could, but it would indeed be bad practice. No worries though!

PR #347, opened 2017-01-01
Fix tests of jUnit5 adapter

SimY4:
Aparently no tests were executed during gradle build for junit5 tests (my bad, sorry). At this point junit5 tests require sepcial gradle plugin for tests discovery. I've added that plugin and fixed couple of issues with current test scenarios.

### DIFF
--- a/subprojects/testfx-junit5/src/test/java/org/testfx/cases/acceptance/ApplicationLaunchTest.java
+++ b/subprojects/testfx-junit5/src/test/java/org/testfx/cases/acceptance/ApplicationLaunchTest.java
@@ -32,13 +32,13 @@ import org.testfx.framework.junit5.ApplicationTest;
 import static org.testfx.api.FxAssert.verifyThat;
 import static org.testfx.matcher.base.NodeMatchers.hasText;
 
-class ApplicationLaunchTest extends FxRobot {
+public class ApplicationLaunchTest extends FxRobot {
 
     //---------------------------------------------------------------------------------------------
     // FIXTURES.
     //---------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


brcolow:
Hmm..thanks for this. I am confused because this is the first time I have seen the legacy TestFX tests fail in this way:

```
org.loadui.testfx.AssertionsTest > classMethod FAILED
    java.awt.AWTError: Can't connect to X11 window server using ':99.0' as the value of the DISPLAY variable.
```

This has happened twice in a row. However I am quite certain it is not related to this PR as the TestFX legacy project is not altered by this PR.

brcolow:
Are the test failures related to this PR?

SimY4:
Hmm... No, I don't think that they are related. Junit5 adapter have exact copies of jUnit4 adapter tests. Can this be related to recent changes in travis CI?

brcolow:
Hmm this is really quite strange because travis is able to build master as it is now on the new infrastructure:

https://travis-ci.org/TestFX/TestFX/jobs/188453444

So I am not sure what is going on.

brcolow:
Apparently these changes are okay if I skip the testfx-legacy tests from running in the build:

https://github.com/TestFX/TestFX/pull/349

I really don't understand why adding these changes causes the legacy tests to fail.

JordanMartinez:
What if you remove the `KeyAndButtonReleaseTest` since that test does need to be run in a fixed order? Is it possible that that test might be interfering with others?

brcolow:
It failed, but I restarted it and it passed. Something is super flaky...but I think this is an improvement. Besides, using the new Travis CI cron builds even master is intermittently failing.

PR #351, opened 2017-01-13
281: Vulnerability in WaitForAsyncUtils

brcolow:
Test


hastebrot:
Removing `@Test(timeout = 1000)`, is this a good idea?

brcolow:
@hastebrot Indeed this is only a work-in-progress, not really ready to be seen.. I should have added a WIP tag (do we have a WIP tag?). Once the tests are passing I will squash to one commit :).

This PR is originally from Ortner, I am just trying to squash it down to one commit and change some stylistic aspects. Then I will investigate the logic behind it.

brcolow:
I am not sure how I feel about having two public static fields (`printException` and `autoCheckException`). It just seems like an awkward API but I am also not sure of a better solution.

brcolow:
This is probably out of scope for this PR but I wonder if introducing `slf4j` as a dependency would make sense (I am thinking about this because this PR uses `System.err.println` to print exceptions. There are many scattered calls to `System.out.println ` throughout the code base...but I wonder if these should be replaced with slf4j calls.

PR #353, opened 2017-01-15
(refactor) Better failure message for ShortcutKeyTest

brcolow:
This makes the assertion failure a bit more readable.

Currently the shortcut test fails on headed macOS builds which is related to https://bugs.openjdk.java.net/browse/JDK-8154148

Things may be different in 8u122, but that is not yet on any Travis macOS version that I am aware of. Even still, the changes in 8u122 may not fix the bug as VMWare + macOS (which is I guess what Travis uses) is "not supported".


JordanMartinez:
@brcolow Are you ever going to merge these changes? It'd need a rebase, but otherwise, it's still good.

brcolow:
Wasn't sure it was worth it. Thanks for the 👍 , going to merge it then!

JordanMartinez:
It's like the keys pressed comment you made earlier: the change isn't
needed but it makes things that much clearer.

Besides, even with the informed handler code, I still couldn't tell why the
shortcut test wasn't working on a flaky build. This might help bring more
details as to why it's not working.

PR #363, opened 2017-02-01
(feature) Introduce the concept of motion.

### LINKED ISSUE #305: Ctrl-A key combination does not work for things other than selecting text
# Issue

The keystroke combination of Ctrl-A does not work for things 'other' than selecting text.

I am using the following command:

```
FxRobot fxr = new FxRobot();
fxr.push(KeyCode.CONTROL, KeyCode.A);
```

For instance, the application I work with has a dialog box that should open when a shortcut key combination is selected. That combination is Ctrl-A.  It was not my decision to use that keystroke, I am just stuck trying to automate it.

We have several menu items across several menus that have shortcut keys to launch their actions, (usually just opening a dialog box).  The only one that does not work in automation is the one that uses Ctrl-A.

brcolow:
Allow the user to explicitly request a certain type of motion
using the API. Automatically detect when certain types of motion
should be used and if the user has not explicitly requested a type
of motion use the most appropriate one.

For example when moving to a MenuItem the mouse should move vertically
(along the y-axis) first and the horizontally so that adjacent menus
are not activated.

Fixes #305.

### DIFF
--- a/subprojects/testfx-core/src/main/java/org/testfx/api/FxRobot.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/api/FxRobot.java
@@ -44,6 +44,7 @@ import javafx.stage.Window;
 
 import org.hamcrest.Matcher;
 import org.testfx.api.annotation.Unstable;
+import org.testfx.robot.Motion;
 import org.testfx.service.locator.PointLocator;
 import org.testfx.service.query.BoundsQuery;
 import org.testfx.service.query.NodeQuery;
@@ -333,7 +334,7 @@ public class FxRobot implements FxRobotInterface {
         PointLocator pointLocator = context.getPointLocator();
         Pos pointPosition = context.getPointPosition();
         targetWindow(node.getScene().getWindow());
-        return pointLocator.point(node).atPosition(pointPosition);
+        return pointLocator.point(node).onNode(node).atPosition(pointPosition);
     }
 
     @Override
@@ -499,7 +500,7 @@ public class FxRobot implements FxRobotInterface {
         waitForFxEvents();
         return this;
     }
-    
+
     @Override
     public FxRobot interactNoWait(Runnable runnable) {
         waitFor(asyncFx(runnable));
@@ -510,7 +511,7 @@ public class FxRobot implements FxRobotInterface {
         waitFor(asyncFx(callable));
         return this;
     }
-    
+
     @Override
     public FxRobot interrupt() {
         waitForFxEvents();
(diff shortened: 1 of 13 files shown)


hastebrot:
Note: This changes parts of TestFX's public API.

hastebrot:
I like the idea of having different types of motions (like euclidean/direct and manhattan/horizontal/vertical) to solve this problem. I think it's great to customize the movement. 

However I tend to be against changing the method signatures. We might instead maybe introduce a `FxRobot#motionTo()` method. Use case `moveTo("File").motionTo("Export", VERTICAL_FIRST).motionTo("XML format", HORIZONTAL_FIRST").click()`.

hastebrot:
~~~java
private void moveMouseStepwiseBetween(
    Point2D sourcePoint, 
    Point2D targetPoint
    Motion motion
) 
~~~

We could possibly make this even more customizable by changing the method signature to

~~~java
private void moveMouseBetween(
    Point2D sourcePoint, 
    Point2D targetPoint
    Point2DInterpolator interpolator
) 
~~~

hastebrot:
An alternative solution for the menu movement problem is to directly jump to the target position without interpolation.

brcolow:
Thank you very much for taking the time to review this thoughtfully, it is much appreciated.

Given that TestFX 4 is "alpha", IMO we don't need to concern ourselves with changing the API much. However even in this case the API is not changed in the sense that it will break anyone's existing code. In fact, with these changes things will behave *exactly* in the same way as before. The only thing that has changed from a public API perspective is more possible arguments (via overloading) to the `moveTo` and `clickOn` methods.

I thought about taking an approach similar the one you listed above wherein we would have a separate, chainable, `motionTo` (or similarly named method) methods for this behavior. But then I thought about it more and realized two things:

1.) Inherent in the concept of `clickOn("#something")` is that if we are not yet at `#something`, we will need to move to it. Indeed, this is why the `clickOn` implementations call `moveTo` if necessary.
2.) Inherent in the concept of `moveTo("#something")` is the question of "how"? We have two distinct points in the plane A and B and we need to get from one to the other. There are many possible paths and currently we take the Euclidean path. But the fact is that the motion is inherently tied to the `moveTo` concept. They are not logically separate and therefore I don't think they should be separated in the public API.

These two points are to show that `clickTo` is really two distinct parts, one of which is `moveTo` and that `moveTo` is inherently related to motion. When not explicitly specified by the user (meaning they end of using one of the `default` methods that calls the overloaded `moveTo` with `Motion.DEFAULT`) it makes  sense to do the *least* surprising thing. This is open to argument but I think we both agree that what you call the Euclidean motion (and this is a good name for it) is the *least* surprising thing. However it makes sense to make the `moveTo` method *expressive* enough to control the motion and that is why I have added the motion argument to the `moveTo` function - because they are so closely tied together.

I am not inflexible in this regard, but did at least want to put my thoughts on the matter out there for more consideration.

In regards to the second point about the `moveMouseBetween` signature - this is part of the private API and so is much less important to get *right*. I think that making the method more flexible is indeed a good idea though, and will at least make a note of it with a comment in the code.

jibbe:
@hastebrot please merge. +1

brcolow:
I think this is okay to merge. I have thought a lot about this API change and I think using default methods is a good solution to this problem.

PR #382, opened 2017-03-27
[WIP] (refactor) Fix many IntelliJ inspection issues.

brcolow:


hastebrot:
This removes unnecessary `public` keywords in interfaces, and generic type declarations.

It also replaces some of the lambda functions blocks with single lambda functions, or uses method references. Hmm, I'm not a big fan of Java 8 method references.

brcolow:
Okay, if you would like me to revert the method reference changes I can. I think they are more succinct, but that's just my opinion.

PR #383, opened 2017-03-28
(fix) Fix drag and drop test flakiness.

brcolow:
This PR fixes test suite flakiness as well as bumping the gradle and junit5 version, as well as some minor tweaks/fixes. Big thanks to @Ortner for his help with fixing the `WaitForAsyncUtilsTest`!


brcolow:
I accidentally requested a review - this is not ready yet. Sorry about that Benjamin :D

brcolow:
@Ortner I am having trouble understanding how this test is supposed to work:

```java
@Test
public void unhandledExceptionTest() throws Throwable {
    WaitForAsyncUtils.printException=false;
    // given:
    Callable<Void> callable = () -> {
        throw new NullPointerException();
    };
    WaitForAsyncUtils.clearExceptions();
    Future<Void> future = WaitForAsyncUtils.async(callable); //First
    try{ //processing first --> not unhandled
        WaitForAsyncUtils.waitFor(50, MILLISECONDS, future);
    }catch(Throwable t){}


    // expect:
    WaitForAsyncUtils.printException=true;
    try{
        WaitForAsyncUtils.checkException();
    }catch(Throwable e){
        assertTrue("Handled exception not removed from stack",false);
    }
    WaitForAsyncUtils.clearExceptions();
    waitForException(future);
    //verify checkException
    WaitForAsyncUtils.clearExceptions();
    WaitForAsyncUtils.checkException();
    //verify exception in Future
    WaitForAsyncUtils.printException=true;
    waitForThreads(future);
}
```

How is it that the exception from `WaitForAsyncUtils.waitFor(50, MILLISECONDS, future);` is removed from the stack? The exception will be added to the stack when the `ASyncFXCallable` is run on the executorService and thus when running `WaitForAsyncUtils.checkException();` the exception will be popped from the stack and thrown. What am I missing!?

I have made some modifications to the code since you made your PR, but if I understand correctly I don't see how this can have an effect on whether or not this test should pass. This one test has always been flaky, but I don't see how it should ever pass!

Thanks in advance for any time you can lend to this issue.

Here is the current [WaitForAsyncUtils.java](https://github.com/TestFX/TestFX/blob/master/subprojects/testfx-core/src/main/java/org/testfx/util/WaitForAsyncUtils.java) and [WaitForAsyncUtilsTest.java](https://github.com/TestFX/TestFX/blob/master/subprojects/testfx-core/src/test/java/org/testfx/util/WaitForAsyncUtilsTest.java)

brcolow:
@Ortner-CI thanks for the quick response! However, I don't see where/when the exception is removed from the stack inside `waitFor`:

```java
public static <T> T waitFor(long timeout, TimeUnit timeUnit, Future<T> future) throws TimeoutException {
    try {
        return future.get(timeout, timeUnit);
    }
    catch (ExecutionException exception) {
        // if the computation threw an exception.
        throw new RuntimeException(exception.getCause());
    }
    catch (InterruptedException ignore) {
        // if the current thread was interrupted while waiting.
        return null;
    }
}
```

Thanks.

brcolow:
Ah..okay, makes sense. Thank you for the explanation! I didn't connect the dots to understand that it is accomplished by overriding the get method of `Future`. Awesome, I should be all set to get this hammered out now. Thanks again.

brcolow:
@Ortner-CI 

Although the frequency of the failure has been lessened by this PR, I can still trigger a failure about 1 in every 10 times (less if you count over different systems):

```
org.testfx.util.WaitForAsyncUtilsTest > unhandledExceptionTest FAILED
    java.lang.AssertionError: Handled exception not removed from stack
        at org.junit.Assert.fail(Assert.java:88)
        at org.testfx.util.WaitForAsyncUtilsTest.unhandledExceptionTest(WaitForAsyncUtilsTest.java:179)
```

brcolow:
@hastebrot I think one of the last remaining flaky bits is the first test to run in the entire test suite - which I think has something to do with the JavaFX platform not being initialized. I did a little research and thought that gradle's [TestListener](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/testing/TestListener.html) might help us - but I am not sure how we can actually check if the JavaFX platform has been initialized from gradle:

```groovy
def listener = new TestListenerImpl()

test {
    addTestListener(listener);
     testLogging {
        // Blah blah
    }
}

class TestListenerImpl implements TestListener {
    void beforeSuite(TestDescriptor suite) { // how can we wait for JavaFX platform to be initialized here? }
    void afterSuite(TestDescriptor suite, TestResult result) {}
    void beforeTest(TestDescriptor test) {}
    void afterTest(TestDescriptor test, TestResult result) {}
}
```

 Also, I am not sure if we can use a junit [RunListener](http://junit.sourceforge.net/javadoc/org/junit/runner/notification/RunListener.html) to override `testRunStarted` because I am not sure there is a way to wire this to the gradle test runner.

Ortner:
You could try:
``FXToolkit.isFXApplicationThreadRunning()``

This should check, if the Application thread is running. It does not check the state actually.

brcolow:
Thanks for the tip about that method - will be useful. Seems someone had a similar question to me: http://gradle.1045684.n5.nabble.com/junit-listener-from-build-gradle-td5710939.html which was never answered. I experimented with trying to call a static method from the `testfx-core:test` module from the gradle build - but that turned into a cyclic dependency nightmare. Besides annotating every TestFX test with a class-rule annotation, I am not sure how to wire a junit test listener to gradle.

brcolow:
I am just experimenting with what it would look like if we went the route of having every `testfx-core` test class have a `TestFXRule` field. This is still very much a WIP.

brcolow:
The non-allowed to fail configurations have passed like 15 times in a row now - but I think the drag and drop is not yet resolved. Need a failure to happen with the extra `println` statements I added so I can diagnose what's going wrong with `left_to_left` test. However the test suite is in way better shape now :).

brcolow:
I *finally* got the drag-and-drop test to fail with logging (took like 25 tries, lol). I have also in those 25 or so attempts seen failures on the headless macOS build - but I think that may just need an adjustment in the timeout time of the JavaFX initialization in the `TestFXRule`.

Haven't had time to analyze this yet, but posting it here for posterity:

Working:

```
org.testfx.cases.integration.DragAndDropTest > should_drag_and_drop_from_left_to_left STANDARD_OUT
    Left list view before drag:[L1, L2, L3]
    Right list view before drag: [R1, R2, R3]
    pointOfVisibleNode returning: PointQueryBase [position = Point2D [x = 0.5, y = 0.5], offset = Point2D [x = 0.0, y = 0.0] node = ListViewSkin$2@544f8b69[styleClass=cell indexed-cell list-cell]'L3']
    Left list view after drag:[L1, L2, L3]
    Right list view after drag: [R1, R2, R3]
    Inside dropTo: L2
    pointOfVisibleNode returning: PointQueryBase [position = Point2D [x = 0.5, y = 0.5], offset = Point2D [x = 0.0, y = 0.0] node = ListViewSkin$2@5b020bee[styleClass=cell indexed-cell list-cell]'L2']
    Inside dropTo
    Moving to: PointQueryBase [position = Point2D [x = 0.5, y = 0.5], offset = Point2D [x = 0.0, y = 0.0] node = ListViewSkin$2@5b020bee[styleClass=cell indexed-cell list-cell]'L2']
    Calling drop
    Left list view after dropTo:[L1, L2, L3]
    Right list view after dropTo: [R1, R2, R3]
    Left list view after waitForFxEvents:[L1, L2, L3]
    Right list view after waitForFxEvents: [R1, R2, R3]
    leftListView: [L1, L2, L3]
    rightListView: [R1, R2, R3]
```

Failing:

```
org.testfx.cases.integration.DragAndDropTest > should_drag_and_drop_from_left_to_left STANDARD_OUT
    Left list view before drag:[L1, L2, L3]
    Right list view before drag: [R1, R2, R3]
    pointOfVisibleNode returning: PointQueryBase [position = Point2D [x = 0.5, y = 0.5], offset = Point2D [x = 0.0, y = 0.0] node = ListViewSkin$2@760ecba8[styleClass=cell indexed-cell list-cell]'L3']
    Left list view after drag:[L1, L2, L3]
    Right list view after drag: [R1, R2, R3]
    Inside dropTo: L2
    pointOfVisibleNode returning: PointQueryBase [position = Point2D [x = 0.5, y = 0.5], offset = Point2D [x = 0.0, y = 0.0] node = ListViewSkin$2@1ca46d91[styleClass=cell indexed-cell list-cell]'L2']
    Inside dropTo
    Moving to: PointQueryBase [position = Point2D [x = 0.5, y = 0.5], offset = Point2D [x = 0.0, y = 0.0] node = ListViewSkin$2@1ca46d91[styleClass=cell indexed-cell list-cell]'L2']
    Calling drop
    Left list view after dropTo:[L1, L2]
    Right list view after dropTo: [R1, R2, R3]
    Left list view after waitForFxEvents:[L1, L2]
    Right list view after waitForFxEvents: [R1, R2, R3]
    leftListView: [L1, L2]
    rightListView: [R1, R2, R3]
org.testfx.cases.integration.DragAndDropTest > should_drag_and_drop_from_left_to_left FAILED
    java.lang.AssertionError: 
    Expected: iterable containing ["L1", "L2", "L3"]
         but: No item matched: "L3"
        at org.testfx.api.FxAssert.verifyThatImpl(FxAssert.java:150)
        at org.testfx.api.FxAssert.verifyThat(FxAssert.java:62)
        at org.testfx.cases.integration.DragAndDropTest.should_drag_and_drop_from_left_to_left(DragAndDropTest.java:139)
```

PR #404, opened 2017-07-21
(feature) Add Spock variant of TestFX

JordanMartinez:
Using `providedCompile` for the groovy dependency fails the build with a `java.lang.ExceptionInInitializerError`. So, I'm using `compile`. Any idea how to get around this?

Other than that, every variant of TestFX (spock, junit, junit5, legacy) are failing their `Application[Start/Launch][Test/Spec]`'s `should_click_on_button` test due to an NPE from the node query ".button".

Also, I realized there's a way to write the `KeyAndButtonReleaseTest` without needing to use `@FixedMethodOrder`. We can fix the JUnit 5 test to use the new approach since it doesn't have JUnit 4's annotation yet.

### DIFF
--- a/subprojects/testfx-spock/src/main/groovy/org/testfx/framework/spock/ApplicationAdapter.groovy
+++ b/subprojects/testfx-spock/src/main/groovy/org/testfx/framework/spock/ApplicationAdapter.groovy
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2013-2014 SmartBear Software
+ * Copyright 2014-2017 The TestFX Contributors
+ *
+ * Licensed under the EUPL, Version 1.1 or - as soon they will be approved by the
+ * European Commission - subsequent versions of the EUPL (the "Licence"); You may
+ * not use this work except in compliance with the Licence.
+ *
+ * You may obtain a copy of the Licence at:
+ * http://ec.europa.eu/idabc/eupl.html
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the Licence is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the Licence for the
+ * specific language governing permissions and limitations under the Licence.
+ */
+package org.testfx.framework.spock;
+
+import javafx.application.Application;
+import javafx.stage.Stage;
+
+public final class ApplicationAdapter extends Application implements ApplicationFixture {
+
+    //---------------------------------------------------------------------------------------------
+    // PRIVATE FIELDS.
+    //---------------------------------------------------------------------------------------------
+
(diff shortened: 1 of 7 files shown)


brcolow:
I will be away for a couple more weeks - really sorry to have not seen this and responded earlier. Will take a look as soon as I can - same with your other open PR. Thanks so much.

JordanMartinez:
No worries! Whenever you get to it.

JordanMartinez:
Are the failed tests in Java 8 builds flaky? When was Java 9 build added?

brcolow:
Yes the failed test is indeed flaky. I restarted Travis and it passed...AppVeyor is still having issues. But it has nothing to do with your work here. Need to address the test flakiness ASAP, it really defeats the purpose of CI.

JordanMartinez:
> Need to address the test flakiness ASAP, it really defeats the purpose of CI.

Aye. Do we have a complete list of the flaky tests? Do we have any indication as to what's making them flaky?

brcolow:
I have not really tabulated which are flaky, but I believe you can browse the build history of Travis and AppVeyor and easily compile such a list. As to the second question I haven't a clue, yet.

brcolow:
All of the new tests pass CI, which is pretty awesome. This can be merged anytime and is a 👍 from me. @hastebrot do you have anything to add or shall we merge as-is?

hastebrot:
This looks very good.

JordanMartinez:
Before merging, any comments on what I mentioned before?

> Using `providedCompile` for the groovy dependency fails the build with a `java.lang.ExceptionInInitializerError`. So, I'm using `compile`. Any idea how to get around this?

brcolow:
I am not sure what the "correct" gradle modifier is for groovy dependency. Does using `compile` instead of `providedCompile` mean that even if people don't use the spock sub-project the dependency will be compiled in? Or is that not the problem the `providedCompile` vs `compile` is supposed to solve?

JordanMartinez:
@brcolow [This issue may be what we need to do](https://discuss.gradle.org/t/providedcompile-for-groovy/6989/7)

JordanMartinez:
compile = the groovy library is added as a dependency to `testfx-spock`
providedComiple = the groovy library is used to compile things, but it's not included in `testfx-spock`

In short, providedCompile is what we'd want to use.

JordanMartinez:
Lol... I tried the build without even including the Groovy dependency and things started up fine (tests still failed, but the fact that they compiled and ran is the key point here). I've updated this PR to remove the Groovy dependency.

brcolow:
Awesome. Thank you for resolving that. Going to merge this and we will work on the test flakiness going forward. Great work.

JordanMartinez:
The tests that failed were all flaky except for the JDK 8 build. It seems like there was an error with spock reports, but I'm not sure. If need be, that dependency can be removed.

PR #405, opened 2017-07-21
(chore) Test key/button release post-test via order-agnostic approach

JordanMartinez:


brcolow:
For junit 4 the tests must remain `public` methods:

```
org.testfx.framework.junit.KeyAndButtonReleaseTest > initializationError FAILED
    java.lang.Exception: Method pressed_buttons_are_released_at_test_end() should be public
org.testfx.framework.junit.KeyAndButtonReleaseTest > initializationError STARTED
Gradle Test Executor 4 finished executing tests.
org.testfx.framework.junit.KeyAndButtonReleaseTest > initializationError FAILED
    java.lang.Exception: Method pressed_keys_are_released_at_test_end() should be public
```

JordanMartinez:
Updated PR to account for your comments

brcolow:
I am no longer certain that these changes that the tests verify that pressed mouse buttons and pressed keys, if forgotten to be released after a test, make sure that a *new* test is started with a clean state.
`internalAfter` explicitly clears what is set, so what are we testing here now?

JordanMartinez:
Sorry but could you reword your comment. It's hard to understand you.

brcolow:
The scenario that we were testing for originally with `KeyAndButtonReleaseTest.java` is that if a developer using TestFX forgot to release a mouse button or key in the "cleanup" portion of a test then we would do it for them so that when a new test begins no keys or mouse buttons are in a pressed state. With these changes since we are explicitly calling `internalAfter()` which clears keys and mouse buttons, what are we testing now?

brcolow:
Would love to get rid of forcing the test order as well, but I don't see how it's possible while retaining the same semantics.

JordanMartinez:
Yeah... Maybe that will change in the future for JUnit 5?

Either way, I'd say this PR doesn't address the issue.

PR #412, opened 2017-08-18
(refactor) Make ApplicationTest#start() not abstract

AustinShalit:
There are times where I override start and it is just a no-op because I only need the JavaFX Application thread to exist.  This makes it so I do not _need_ to override the start method.


brcolow:
Makes sense. Thank you.

Issue #411, opened 2017-08-17
Deflake flaky tests


JordanMartinez:
As stated in #404, compile a list of flaky tests and get rid of them. That PR could have been merged much earlier but wasn't due to flaky tests. Removing this bottleneck should speed up development for future PRs.

JordanMartinez:
## Windows (only appears to be 1 flaky test)
- org.testfx.cases.acceptance.FxAssertBasicTest > button_has_label FAILED
   - [Headed - master 310](https://ci.appveyor.com/project/testfx/testfx/build/master%20310/job/35svwog76qlbpy6m#L1082)
   - [Headless - master 309](https://ci.appveyor.com/project/testfx/testfx/build/master%20309/job/8ri3u6qdxeqphhba#L1082)
````
    java.lang.AssertionError: 
    Expected: Node has text "clicked!"
         but: was <Button[id=button, styleClass=root button]'click me!'>
        at org.testfx.api.FxAssert.verifyThatImpl(FxAssert.java:150)
        at org.testfx.api.FxAssert.verifyThat(FxAssert.java:84)
        at org.testfx.cases.acceptance.FxAssertBasicTest.button_has_label(FxAssertBasicTest.java:99)
````

JordanMartinez:
## OSX (1 Full Headless; 7 Partial Headless)

Where "___ Headless" is based on the environment variables
````json
# Full Headless
"env": "_JAVA_OPTIONS=\"-Djava.awt.headless=true -Dtestfx.robot=glass -Dtestfx.headless=true -Dprism.order=sw -Dprism.text=t2k -Dtestfx.setup.timeout=2500\""
# Partial Headless
"env": "OSX_HEADLESS=false _JAVA_OPTIONS=\"-Dprism.order=sw\""
````
- Full Headless - [org.testfx.service.finder.impl.WindowFinderImplTest STANDARD_ERROR](https://travis-ci.org/TestFX/TestFX/jobs/262166281#L4662)
````
    Exception in thread "JavaFX Application Thread" java.lang.NullPointerException
    	at com.sun.javafx.tk.quantum.PaintCollector.renderAll(PaintCollector.java:402)
    	at com.sun.javafx.tk.quantum.QuantumToolkit.pulse(QuantumToolkit.java:511)
    	at com.sun.javafx.tk.quantum.QuantumToolkit.pulse(QuantumToolkit.java:490)
[...]

PR #413, opened 2017-08-20
[chore] Split Travis apt packages for *nix by JDK.

brcolow:
As suggested by @JordanMartinez in https://github.com/TestFX/TestFX/issues/411


JordanMartinez:
Glad this worked!

PR #414, opened 2017-08-21
Refactor spock

JordanMartinez:
Continuing what was started in #404, tests whether removing the Spock reports runtime dependency affects the success of a previously-failing test on the allow-fail JDK 8 build


JordanMartinez:
This proves that removing Spock reports gets rid of the error on the allow-fail Java 8 build. Perhaps the original error was flaky itself, but who knows?

brcolow:
Thanks!

PR #421, opened 2017-08-23
[RFC] Extract failed test debugger code into its own utiltiy class

### LINKED ISSUE #420: Improve failed test's AssertionError message mapping
As discussed in the Gitter Chat, a few other ideas for improving #418 are:

- extract the third argument (FxRobot) in verifyThat into a DebugUtils class. Then, one could write verifyThat(node, matcher, DebugUtils.informedErrorHandler(thisAsFxRobot)
- in the above DebugUtils class, separate items into functions that can be composed, so that one can quickly add only the things they need to see (e.g. keys pressed, buttons pressed, etc.)
- change the mapping function to `UnaryOperator<String>`, so that the developer can run additional things within the error handler without forgetting to return a new AssertionError (they may return the original one). Then, the error handler code would look like:

````java
try {
    matcher.match(node);
} catch (AssertionError error) {
    throw`new AssertionError(mapper.apply(error.getMessage));
}
````

### PR DESCRIPTION
**Pull request by JordanMartinez**
Resolves #420

### DIFF
--- a/subprojects/testfx-core/src/main/java/org/testfx/api/FxAssert.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/api/FxAssert.java
@@ -35,6 +35,11 @@ import org.testfx.matcher.base.GeneralMatchers;
 import org.testfx.service.finder.NodeFinder;
 import org.testfx.service.query.NodeQuery;
 
+/**
+ * All TestFX tests should use {@link #verifyThat(Node, Matcher, Function)} when writing tests,
+ * so that the developer can use {@link org.testfx.util.DebugUtils} to provide additional info
+ * as to why a failed test failed.
+ */
 @Unstable(reason = "method signatures need fine-tuning")
 public final class FxAssert {
 
@@ -212,6 +217,8 @@ public final class FxAssert {
 
     /**
      * Allow developer to debug a failed test (e.g. the state of the stage or node, which keys were pressed, etc.)
+     *
+     * @see org.testfx.util.DebugUtils
      */
     private static <T> void verifyThatImpl(String reason, T value, Matcher<? super T> matcher,
                                            Function<StringBuilder, StringBuilder> errorMessageMapper) {
--- a/subprojects/testfx-core/src/main/java/org/testfx/api/FxAssert.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/api/FxAssert.java
@@ -20,15 +20,12 @@ import java.nio.file.Paths;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.Set;
+import java.util.function.Function;
 import java.util.function.Predicate;
-import java.util.function.UnaryOperator;
 
-import javafx.event.Event;
 import javafx.geometry.Rectangle2D;
 import javafx.scene.Node;
 import javafx.scene.image.Image;
-import javafx.scene.input.KeyCode;
-import javafx.scene.input.MouseButton;
 import javafx.stage.Screen;
 
 import org.hamcrest.Matcher;
@@ -64,21 +61,14 @@ public final class FxAssert {
     @Unstable(reason = "is missing apidocs; might be removed due to complications with generics")
     public static <T> void verifyThat(T value,
                                       Matcher<? super T> matcher) {
-        verifyThatImpl(emptyReason(), value, matcher, defaultErrorHandler());
+        verifyThatImpl(emptyReason(), value, matcher, Function.identity());
     }
 
     @Unstable(reason = "is missing apidocs; might be removed due to complications with generics")
     public static <T> void verifyThat(T value,
                                       Matcher<? super T> matcher,
-                                      UnaryOperator<AssertionError> errorHandler) {
-        verifyThatImpl(emptyReason(), value, matcher, errorHandler);
-    }
-
-    @Unstable(reason = "is missing apidocs; might be removed due to complications with generics")
-    public static <T> void verifyThat(T value,
-                                      Matcher<? super T> matcher,
-                                      FxRobot robot) {
-        verifyThatImpl(emptyReason(), value, matcher, robot);
+                                      Function<StringBuilder, StringBuilder> errorMessageMapper) {
+        verifyThatImpl(emptyReason(), value, matcher, errorMessageMapper);
     }
 
     // ASSERTIONS: {NODE, NODES} + MATCHER.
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: subprojects/testfx-core/src/main/java/org/testfx/api/FxAssert.java [hunk 1]
```diff
@@ -38,6 +35,11 @@
 import org.testfx.service.finder.NodeFinder;
 import org.testfx.service.query.NodeQuery;
 
+/**
+ * All TestFX tests should use {@link #verifyThat(Node, Matcher, Function)} when writing tests,
```
brcolow: Maybe we should be explicit and say "All internal TestFX tests"

JordanMartinez: Why? I really meant all tests should use the `verifyThat(Node, Matcher)`-related assertions because if something goes wrong, one can quickly add an error handler (type now becomes `verifyThat(Node, Matcher, Function)`) that provides additional info to help one debug it. This applies to both TestFX contributors and users of this test library

JordanMartinez:
Before merging, this PR should include tests that prove the code works....

subprojects/testfx-core/src/main/java/org/testfx/api/FxAssert.java
JordanMartinez: Perhaps I should reword it to explain that better?

subprojects/testfx-core/src/test/java/org/testfx/util/DebugUtilsTest.java
```diff
@@ -0,0 +1,296 @@
...
+import static org.testfx.util.DebugUtils.showFiredEvents;
+import static org.testfx.util.DebugUtils.showKeysPressedAtTestFailure;
+import static org.testfx.util.DebugUtils.showMouseButtonsPressedAtTestFailure;
+
+public class DebugUtilsTest extends TestCaseBase {
+
+    private static final String TAB = " ";
+
+    private static final MouseButton MOUSE_BUTTON = MouseButton.PRIMARY;
+    private static final KeyCode KEY = KeyCode.A;
```
brcolow: I don't think that these fields add anything useful. I understand if the intention is to make it clear that it can be any mouse button or key, but I think it complicates reading the test.

subprojects/testfx-core/src/test/java/org/testfx/util/DebugUtilsTest.java
```diff
@@ -0,0 +1,296 @@
...
+        ) {
+
+            @Override
+            public boolean matches(Object item) {
+                return !replaceBlankPart(item).contains(text);
+            }
+        };
+    }
+
+    private abstract class RemoveBlankMessageMatcher extends BaseMatcher<String> {
```
brcolow: What exactly does this class do? Having a hard time understanding it.

subprojects/testfx-core/src/test/java/org/testfx/util/DebugUtilsTest.java
```diff
@@ -0,0 +1,296 @@
...
+            FxAssert.verifyThat(false, alwaysFail(), errorMessageMapper);
+            throw new AssertionError("Error: verifyThat did not throw an error when it should have.");
+        }
+        catch (AssertionError error) {
+            return error;
+        }
+    }
+
+    private Matcher<Object> alwaysFail() {
+        return new AlwaysFailsMatcher();
```
brcolow: Could this be replaced by returning an anonymous inner class that extends Hamcrest's `IsNot` matcher with `not()` overriden to always return `true`? Would that make things any clearer? Not sure.

subprojects/testfx-core/src/test/java/org/testfx/util/DebugUtilsTest.java
```diff
@@ -0,0 +1,296 @@
...
+        AssertionError error = getThrownErrorPostMapper(showFiredEvents(events, TAB));
+
+        // then:
+        assertThat(error.getMessage(), hasAddedTextThatContains("MouseEvent"));
+        assertThat(error.getMessage(), hasAddedTextNotContaining("ScrollEvent"));
+    }
+
+    private AssertionError getThrownErrorPostMapper(Function<StringBuilder, StringBuilder> errorMessageMapper) {
+        try {
+            FxAssert.verifyThat(false, alwaysFail(), errorMessageMapper);
```
brcolow: Can we static import `verifyThat` here?

subprojects/testfx-core/src/main/java/org/testfx/util/DebugUtils.java
```diff
@@ -0,0 +1,213 @@
...
+ * {@link org.testfx.api.FxAssert#verifyThat(Node, Matcher)} or its related methods.
+ *
+ * <p>
+ *     Quickly chain things together using {@link #compose(Function[])} or use standard handlers like
+ *     {@link #informedErrorMessage(String, boolean, FxRobot, boolean, boolean)}. Any {@code tab} parameters in methods
+ *     are there to specify what to insert as a tab. This could be a tab or a number of spaces. Default tab
+ *     value is three spaces (e.g. {@code "   "}).
+ * </p>
+ */
+public final class DebugUtils {
```
brcolow: Should we maybe say "are there to specify what text is inserted for spacing purposes. This defaults to blah blah". What I mean is instead of explicitly calling it "tab", maybe rename to `spacing`?

subprojects/testfx-core/src/test/java/org/testfx/util/DebugUtilsTest.java
JordanMartinez: If the test's assertion should fail, I need to release the mouse button or key that was pressed. Rather than defining that in two separate places that can change, I define it once here and refer to it twice (1x in test, 1x in cleanup).

subprojects/testfx-core/src/test/java/org/testfx/util/DebugUtilsTest.java
JordanMartinez: Sure thing!

subprojects/testfx-core/src/test/java/org/testfx/util/DebugUtilsTest.java
JordanMartinez: I'm not familiar with Hamcrest and their matchers, so I ran my own after looking at the ones created in TestFX itself.

It probably would make things clearer if a developer reading through these test knows Hamcrest well.

subprojects/testfx-core/src/test/java/org/testfx/util/DebugUtilsTest.java
JordanMartinez: It removes the blank part (i.e. the the quotes part in `"Expected: <expectedValue>\n\tbut: "<actualValue>") from the returned error's message since I can't get rid of that myself but still need a better way to determine whether a test worked or not. 

I don't think it's the best designed class because I initially made it do something else before reworking it a bit. The ">>>message<<<" was necessary for debugging my tests until I replaced newline characters with the "\n" string. I'm not sure if it's needed anymore.

subprojects/testfx-core/src/main/java/org/testfx/util/DebugUtils.java
JordanMartinez: Perhaps a better name would be "indent"? 

"Tab" is good in that that is what it's used for, indenting stuff. However, it's bad in that one may assume one needs to pass in "\t". "Spacing" isn't good either, though it fixes the problem with "tab," in that it doesn't communicate the indenting nature of it all. What are we spacing? Space between the current added text and the next one?

subprojects/testfx-core/src/test/java/org/testfx/util/DebugUtilsTest.java
JordanMartinez: I looked into it. `IsNot` requires another Matcher to be passed into it. So, I think it's just easier to leave it as is. Besides, it also overrides the `describeMismatch()` to remove the "was" that gets appended to the description.

subprojects/testfx-core/src/main/java/org/testfx/util/DebugUtils.java
brcolow: Indent works better IMO

subprojects/testfx-core/src/test/java/org/testfx/util/DebugUtilsTest.java
JordanMartinez: I just realized that this isn't actually needed. Why? Because there's a much easier way to remove the blank part by taking any function `f` and calling `f.compose(sb -> new StringBuilder())`

brcolow:
Awesome. Thanks @JordanMartinez

Issue #306, opened 2016-10-01
Screenshot feature removed in 4.0.4


adamish:
The screenshot on failure functionality in 3.x appears to have been removed in 4.0.4.

Could this be restored?

hastebrot:
This is planned for the next release.

brcolow:
I am not sure but I think that to implement this feature for the junit5 provider we will need https://github.com/junit-team/junit5/issues/618 to be resolved. I will work on adding this functionality to the junit4 provider either in https://github.com/TestFX/TestFX/pull/355 where the capture functionality was added or I will extract out the capture functionality from 355.

brcolow:
I think that screenshot capture should be moved out of `testfx-core` and into the junit sub-projects. This way we can get better naming for the saved images and even report the failures with their respective screenshots by utilizing reporter functionality. As I stated I believe `junit5` is not quite yet ready to support this (however I could be wrong about that and if so, yay) but `junit4` can certainly support this with its' `TestWatcher` interface. I will work on this and open a PR very soon.

hastebrot:
>screenshot capture should be moved out of testfx-core and into the junit sub-projects

Sounds good to me. :smiley:

raniejade:
Is there a plan to re-enable this?

```java
private static <T> void verifyThatImpl(String reason,
                                           T value,
                                           Matcher<? super T> matcher) {
        try {
            MatcherAssert.assertThat(reason, value, matcher);
        }
        catch (AssertionError error) {
            // TODO: make error capture and assertion throw more reliable.
            // captureErrorScreenshot();
            throw new AssertionError(error.getMessage());
        }
    }
```

JordanMartinez:
@brcolow Should this feature be re-added to `DebugUtils`? One could execute the `captureErrorScreenshot()` using `DebugUtils#runCode`, correct?

brcolow:
Can we add the auto-capture behavior to `informedErrorHandler()`? It could print out the location of the screenshot (not super useful for CI without further infrastructure in place such as uploading failure screenshots to some s3 bucket or something) but definitely useful when running the our test suite (or developers using TestFX to run their tests locally) locally.

adamish:
I ended writing a simple JUnit @Rule that catches any exception and takes screenshot. I found this more useful as my tests involve fx matchers, various junit/hamcrest asserts and mockito. This approach works in all cases...

brcolow:
[...]

PR #422, opened 2017-08-25
[RFC] (feature) Re-add capture support by moving it to DebugUtils

JordanMartinez:
Implements the idea I mentioned [in my comment in #306](https://github.com/TestFX/TestFX/issues/306#issuecomment-324820682)

### DIFF
--- a/subprojects/testfx-core/src/main/java/org/testfx/util/DebugUtils.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/util/DebugUtils.java
@@ -23,6 +23,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.function.Supplier;
+import java.util.stream.Stream;
 
 import javafx.event.Event;
 import javafx.geometry.Bounds;
@@ -69,6 +70,10 @@ public final class DebugUtils {
         throw new IllegalStateException("Cannot initialize DebugUtils");
     }
 
+    //---------------------------------------------------------------------------------------------
+    // COMPOSITION AND CREATION SUPPORT.
+    //---------------------------------------------------------------------------------------------
+
     /**
      * Composes multiple functions together into one. The ones to run first should be the first in the array,
      * and the ones to run last should be at the end of the array.
@@ -99,10 +104,6 @@ public final class DebugUtils {
         };
     }
 
-    //---------------------------------------------------------------------------------------------
-    // ERROR MESSAGE SUPPORT.
-    //---------------------------------------------------------------------------------------------
-
     /**
      * Inserts a header on a newline; useful for specifying a section in the message
      */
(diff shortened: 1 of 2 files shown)


JordanMartinez:
Can I not write tests for this? I can only imagine such tests will be hard to write and the functional programming style I am using is easy to prove and understand. #lazy

subprojects/testfx-core/src/main/java/org/testfx/util/DebugUtils.java
```diff
@@ -16,31 +16,48 @@
...
+ * a test fails using {@link org.testfx.api.FxAssert#verifyThat(Node, Matcher)} or its related methods.
  *
  * <p>
  *     Quickly chain things together using {@link #compose(Function[])} or use standard handlers like
  *     {@link #informedErrorMessage(String, boolean, FxRobot, boolean, boolean)}. Any {@code indent} parameters
  *     in methods are there to specify what to insert to offset values. Default indent value is three spaces
  *     (e.g. {@code "   "}).
  * </p>
+ * <p>
+ *     When a test fails, an image of the it can be captured and saved to a PNG file. Supports the size
```
brcolow: "When a test fails, an image of the it" => "When a test fails a screenshot of the scene can be captured and saved to a PNG file." Not sure about the rewording, but there is a typo with "of the it".

subprojects/testfx-core/src/main/java/org/testfx/util/DebugUtils.java
```diff
@@ -16,31 +16,48 @@
...
+ *     of a screenshot, a window, an area of the screen (i.e. bounds), or an individual node. Use the
```
brcolow: Also typo with "Supports the of a screenshot" - I assume you accidentley left out "taking" as in "Supports the taking of a screenshot"

subprojects/testfx-core/src/main/java/org/testfx/util/DebugUtils.java
JordanMartinez: Thanks for the catch on the typos. It was caused by editing it after writing something else originally. I was having trouble deciding how to word things, and I like you wording much better.

JordanMartinez:
Originally posted in #306 by @brcolow:

> Can we add the auto-capture behavior to `informedErrorHandler()`? It could print out the location of the screenshot (not super useful for CI without further infrastructure in place such as uploading failure screenshots to some s3 bucket or something) but definitely useful when running the our test suite (or developers using TestFX to run their tests locally) locally.

and...

>  We should make the screenshot feature of `verifyThat` as easy as possible to toggle (should it default to off?) on or off,

I think adding auto-capture behavior to `informedErrorHandler()` makes sense. However, what would we put for the test name? We could use `"testfx-test - " + atomicIntegerPhotoNumber.getAndIncrement() + ".png"` for the file path, but how does one know which image corresponds to which test? Is there an easier way to figure out the test name (say through the Thread's stack)?

I'd say default it to on. If the test doesn't fail, the developer doesn't care. If it does, then this gives them all the info they need. Most will be grateful that such handling is already provided as default. However, if there are some who don't like it, the DebugUtils API allows them to easily roll their own mapper. If we leave it off, I fear the capture support will only be known by those familiar with the code and its updates or only be discovered after much frustration at a failed test. In other words, "convention over configuration."

subprojects/testfx-core/src/main/java/org/testfx/util/DebugUtils.java
JordanMartinez: On second thought, I think I'll just remove the typo "the." The concept I want to introduce/begin with in the paragraph's first sentence is the idea of capturing an image of the test and later clarify what kinds of images can be captured. I don't want to say screenshot initially because it starts to leads people astray as to what is supported.

subprojects/testfx-core/src/main/java/org/testfx/util/DebugUtils.java
JordanMartinez: Actually, that was supposed to indicate what sizes of images it could take. I've reworded it to better clarify that.

JordanMartinez:
> However, what would we put for the test name? We could use "testfx-test - " + atomicIntegerPhotoNumber.getAndIncrement() + ".png" for the file path, but how does one know which image corresponds to which test?

Perhaps the file name should be appended to the error message itself? Then, when the error is thrown, it'll tell the developer the path to look to find the corresponding screenshot.

JordanMartinez:
The most recent push doesn't build on Windows because it failed to download Gradle. Any idea why?

brcolow:
I think they are just having connectivity issues, because I saw that with the gradle 4.1 PR as well, I restarted it.

JordanMartinez:
Looks like this is good to go. The Windows build failed only due to the flaky tests.

brcolow:
Great work @JordanMartinez - thanks a lot.

JordanMartinez:
Woot!

PR #426, opened 2017-08-30
(chore) Remove Travis macOS "partial headless" build.

brcolow:


JordanMartinez:
Just a thought, but shouldn't we also update the example `travis.yml` file in the wiki to explain the partial headless issues? Or should we also remove the partial headless build from that example file as well?

PR #438, opened 2017-10-07
Flaky test: org.testfx.cases.acceptance.FxAssertBasicTest > button_has_label 

Oliver-Loeffler:
When the action is consumed, then on Window and MacOS the test will pass repeatedly.
I couldnt reproduce any failures to this test after the change.

However, I'm not sure yet if this is the right way, so I'm happy about feedback.
Thanks.

### DIFF
--- a/subprojects/testfx-core/src/test/java/org/testfx/cases/acceptance/FxAssertBasicTest.java
+++ b/subprojects/testfx-core/src/test/java/org/testfx/cases/acceptance/FxAssertBasicTest.java
@@ -17,10 +17,8 @@
 package org.testfx.cases.acceptance;
 
 import java.util.concurrent.ExecutionException;
-import java.util.concurrent.FutureTask;
 
 import javafx.application.Application;
-import javafx.application.Platform;
 import javafx.scene.Scene;
 import javafx.scene.control.Button;
 import javafx.stage.Stage;
@@ -96,10 +94,10 @@ public class FxAssertBasicTest extends TestCaseBase {
     }
 
     @Test
-    public void button_has_label() {
+    public void button_has_label() throws InterruptedException, ExecutionException {
         // when:
         clickOn("#button");
-
+        
         // then:
         verifyThat("#button", hasText("clicked!"), informedErrorMessage(this));
     }
@@ -124,28 +122,13 @@ public class FxAssertBasicTest extends TestCaseBase {
         public void start(Stage stage) {
             Button button = new Button("click me!");
             button.setId("button");
-            button.setOnAction(actionEvent -> {
-                invokeAndWait(() -> button.setText("clicked!")); });
+            button.setOnAction(action -> { 
+                button.setText("clicked!"); 
+                action.consume(); });
             Scene scene = new Scene(button, 600, 400);
             stage.setScene(scene);
             stage.setTitle(getClass().getSimpleName());
             stage.show();
-        }
-        
-        public void invokeAndWait(Runnable r) {
-            if (Platform.isFxApplicationThread()) {
-                r.run();
-            } else {
-                System.out.println("Not in JavaFX GUI Thread");
-                FutureTask<Void> task = new FutureTask<>(r, null);
-                Platform.runLater(task);
-                try {
-                    task.get();
-                } 
-                catch (InterruptedException | ExecutionException e) {
-                    throw new RuntimeException(e);
-                }
-            }
         }
     }
 }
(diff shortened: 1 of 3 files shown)


Oliver-Loeffler:
Interestingly following approaches do work within the test itself, even when the action event is not consumed.
```
        // works
        clickOn("#button").release(MouseButton.PRIMARY);
```

Following procedure does not work when action event is not consumed:
```
        // works not
        clickOn("#button").moveBy(2,2);
        clickOn("#button").clickOn("#button");
```

JordanMartinez:
Thats... weird...

Oliver-Loeffler:
I'm learning.
So I now understand and found, that calling `ActionEvent.consume()` does not help at all - it should not and it does not. 

Running `gradle clean check` on command line still exposes the `FxAssertBasicTest.button_has_label()` failing repeatedly. So this is really the flakyness you were talking about. I also learned that I have to keep track of my window arrangements :-)

So I continued to play with the framework and tested from within IntelliJ IDEA and Eclipse Oxygen. Well, both IDEs run the tests successfully on MacOS and Windows. The condition, where the FxAssertBasicTest fails is the following:

 1. Eclipse IDE running in background
 2. CMD window in front
 3. executing `gradle clean check` within CMD
    * `FxAssertBasicTest.button_has_label()` fails immediately > then the first test stage becomes visible
    * Before first stage appears visible on screen, the Java window icon starts to blink in task bar

Cool is, that failing tests automatically produce screen shots and what does the first test (failing) screen shot show? No visible stage but an icon indicating activity - so stage is behind the CMD window.
missingstage - testfx-test - 1

If I understand correctly, the FxRobot controls system inputs, so consequently the mouse would click into the CMD window instead on the button on stage. Please correct me if I'm wrong here. 

Thus, I tried to force the stage to appear on top of all other windows using `stage.setAlwaysOnTop(true);` within the DemoApplication. There it works fine. But this has one disadvantage, if multiple stages would exist, the primary one would be above all others.
Therefore I also tested `stage.toFront();`. Interestingly `stage.toFront()` does not help as the window stays hidden. The only way to ensure that the click goes on the button is to set `stage.setAlwaysOnTop(true);`.

Are you able to reproduce this @JordanMartinez ?

Oliver-Loeffler:
Is there a method which reliably will bring a JavaFX stage to front on demand? `toFront()` does not seem to be reliable, or I'm using it the wrong way.

brcolow:
Did the line endings get changed or something? Why is the diff so large for ApplicationLaunchTest.java? Other than that always setting the Stage to be on top makes sense to me.

Oliver-Loeffler:
I will change the formatting, for ApplicationLaunchTest.java the wrong code formatting was applied.

Oliver-Loeffler:
Yes, formatting was broken but is now corrected. The diff is now as small as it should be.

brcolow:
Merged (I squashed all commits into a single one) - thanks very much for tracking this down!

brcolow:
@Oliver-Loeffler Still seeing a failure with button_is_not_null - was that supposed to be address by this PR or only button_has_label? Failure [here](https://travis-ci.org/TestFX/TestFX/jobs/286298410#L1736)

Oliver-Loeffler:
Did not know that button_is_not_null is failing. I'll look into it.

Oliver-Loeffler:
That is an interesting interaction. The button_has_label fails when the stage just does not come in front at the right time. Forcing it into front permanently, so making it modal, will cause a timeout for button_is_not_null, even locally on macOS that happens.

So I modified code like the following:

```java
@Override
public void start(Stage stage) {
    Button button = new Button("click me!");
    
    [...]
   
    if (isWindows()) { 
        stage.setAlwaysOnTop(true);
    }
}

private boolean isWindows() {
    return System.getProperty("os.name").toLowerCase().indexOf("win") >= 0;
}
```

Yet, I'm not happy with it as I just don't know how you generally deal with such little differences in platform behavior, I have to look for some tests where you actually do this - the I'd follow the same style & concept.

And, actually toggling stage modality at the beginning of the test is enough to overcome the Windows behavior, there is no need to keep the stage modal.

To avoid such fixing & breaking at same time, I've setup also my own Travis build to test this before.

brcolow:
I wonder if it's just because it is the first test that executes and TestFX has not been initialized fully yet? I am happy with changes that are platform-specific to the tests as long as it's not some sort of hack. Using the JavaFX API to make the Stage be on top (which we are kind of assuming it is in our tests) is okay IMO.

JordanMartinez:
@brcolow are you suggesting increasing the amount of time before the first test is run?

brcolow:
@JordanMartinez If that's actually the problem and it fixes it - sure :)

Issue #441, opened 2017-10-15
Enhancement Request: Include OSGi headers in JAR Manifest


jtkb:
I've been developing my own framework to allow JavaFX to take advantage of OSGi features ([here](https://github.com/jtkb/osgifx) and have been using TestFx to write some simple integration tests. Unfortunately the integration tests fail intermittently due to a class loading issue [described here](https://github.com/jtkb/osgifx/issues/1). These failures can be attributed to the TestFx jar file not being wrapped and resolved prior to the test commencing (some of the time).

One way to resolve this issue would be to include OSGi headers in the JAR manifest and introduce a 'dummy service'  that is registered once the TestFx bundle is activated. Test class could then request this service to be injected and hence would not execute until this dependency was available.

I have experimented by modifying TestFx to be built in such a way, [here](https://github.com/jtkb/TestFX/tree/feature/move_to_maven). (I had to move to Maven as I don't know Gradle and I find it too complicated for my needs).

The main change I have made is to the [testfx-core](https://github.com/jtkb/TestFX/tree/feature/move_to_maven/testfx-corel) where I have included a new sub-package ```org.testfx.osgi```. Also the [pom.xml](https://github.com/jtkb/TestFX/blob/feature/move_to_maven/testfx-core/pom.xmll) shows the maven bundle plugin settings I have used.

An example of a test using this change is [here](https://github.com/jtkb/osgifx/blob/bugfix/issue_1_testfx_failure/it/tests/src/test/java/com/javatechnics/osgifx/boot/TFxTest.java)

Is there any possibility such a change could be introduced into TestFx? There would be a build dependency on an OSGi framework bundle but wouldn't be needed otherwise except in an OSGi framework as no existed TestFX classes are modified to import OSGi packages.

brcolow:
I would be open to merging a PR that adds such support - but it would need to use gradle as that's the build system TestFX uses.

jtkb:
Sure thing. I think I need to modify the Gradle files in testfx-core, -internal-java8 and internal-java9 to build with the OSGi or bnd plugin so I will experiment with that. I also note issue #443 and I made a comment on there.

PR #446, opened 2017-10-29
(enhancement) Issue #441 - OSGi Headers in Manifest

jtkb:
Modifies the Gradle build files in -core, -java8 and -java9 to build OSGi compatible bundles and introduces OSGi specific class 'Activator'.

-core bundle acts as a Host to either -java8 or -java9 (Fragment) bundles and one of the Fragments must be installed during test for -core to become active.

Dependency on findbugs.annotations caused issues as JSR305 is not yet official and shares package space, javax.annotations,  with annotations 1.2. Explicitly forced version dependency but there can be issues with Guava which has an optional dependency on java.annotations which unless handled carefully can be resolved to annotations 1.2.

No existing TestFx code has been modified and only a single OSGi specific Activator class has been introduced. This provides an injectable service which can be used to prevent any tests (e.g. Pax-Exam) from running until TestFx has been resolved.

### DIFF
--- a/subprojects/testfx-core/testfx-core.gradle
+++ b/subprojects/testfx-core/testfx-core.gradle
@@ -28,6 +28,7 @@ dependencies {
     compile "com.google.guava:guava:21.0"
     compile "org.hamcrest:hamcrest-core:1.3"
     compile 'com.google.code.findbugs:annotations:3.0.1u2'
+    compile 'org.osgi:org.osgi.core:6.0.0'
 
     testCompile "junit:junit:4.12"
     testCompile "org.mockito:mockito-core:2.7.22"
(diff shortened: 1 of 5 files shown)


subprojects/testfx-core/testfx-core.gradle
```diff
@@ -28,6 +28,7 @@ dependencies {
     compile "com.google.guava:guava:21.0"
     compile "org.hamcrest:hamcrest-core:1.3"
     compile 'com.google.code.findbugs:annotations:3.0.1u2'
+    compile 'org.osgi:org.osgi.core:6.0.0'
```
brcolow: I am not sure, but I wonder if the scope of this dependency should be changed to [compileOnly](https://blog.gradle.org/introducing-compile-only-dependencies). Can you see if in doing so it still works for you? It may not be an appropriate usage but I am trying to make the dependency "more optional".

brcolow:
Thanks very much for stepping up and tackling this. I left a small, nitpicky comment. Let me know what you think otherwise this LGTM.

subprojects/testfx-core/testfx-core.gradle
jtkb: Sure thing. Will update this PR.

brcolow:
Merged. Thanks again!

jtkb:
No problem. I'll create another commit soon to documenting/point to my own repo how TestFx can be used in OSGi container.

PR #450, opened 2017-11-06
(fix) Remove Findbugs annotations.

brcolow:
This is done so that the dreaded jsr305 dependency is not pulled
in transitively for TestFX. This particular dependency has caused many
issues with Java 9 migration because of it using the "javax.annotation"
package. It is easiest for us just to drop the one usage we have of it.

Ping @jtkb - does this break your work on osgi or is it okay?


jtkb:
@brcolow I approve the removal of the dependency on findbugs annotations as it's use of the package 'javax.annotation' doesn't sit nicely in OSGi.

You've removed the explicit import of the package space from findbugs so that's all that is required. Should it break anything for me I'll just make a PR to fix but I don't anticipate any issues.

brcolow:
Thanks very much :)

PR #460, opened 2017-11-17
(enhancement) Add module-info.java to testfx-core.

brcolow:
This is an attempt to begin the process of jigsaw modularization for
Java 9. Currently we are not taking advantage of modules for the
different parts of testfx-core, we are simply making everything one big
module but it's the first step.

hamcrest-core and hamcrest-library were used in the compileTestJava
phase of the build and these two artifacts have a split package (both
use the same org.hamcrest package). Instead of messing about with
--patch-modules and the like we replaced all usages of hamcrest-library
with hamcrest-core CoreMatchers where possible. The only place where
this was not feasible was the "closeTo" matcher, which we copied into
CaptureSupportImplTest wholesale. If it turns out we need this Matcher
across the tests we will use a different method of "fixing" this.

### DIFF
--- a/.ci/after_success.sh
+++ b/.ci/after_success.sh
@@ -6,20 +6,9 @@ JAVA_VER=$(java -version 2>&1 | grep -i version | sed 's/.*version ".*\.\(.*\)\.
 echo "Checking java version: ${JAVA_VER}"
 if [[ "${JAVA_VER}" == 8 ]]; then
   echo "Uploading code coverage results to codecov.io"
-  # Download codecov bash from master branch as codecov.io/bash lags behind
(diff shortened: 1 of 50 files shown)


JordanMartinez:
> lol

Just thought I'd share that your commit message made me laugh :laughing:

brcolow:
Context: "lol" and it's successors were rebased away. Probably will be replaced by "lmao", "rofl", "i quit" before long but all will be rebased away to one commit where it appears I am a UNIX master.

JordanMartinez:
I like the "I quit" one :smile:

JordanMartinez:
Maybe you need to call it a day, take a step back for a bit, and the solution will come to you tomorrow?

PR #462, opened 2017-11-23
Test the impossible.

brcolow:


JordanMartinez:
It actually passed!? Lol...

brcolow:
This was due to a faulty implementation of apply() and statement() in `TestFXRule` - fixed in #461

PR #474, opened 2017-12-09
(chore) Add test failure summary.

brcolow:
This is nice as we don't have to go looking through a large build log
for individual failures, they are always printed at the end.


JordanMartinez:
Kinda makes you wonder why that's not supported by default with the option of disabling it...

Issue #475, opened 2017-12-10
Readme example doesn't work


jcornaz:
Here is a sample test written following the example described on the [readme](https://github.com/TestFX/TestFX/blob/97a50a1d7ecc4e33d431400c34d3e0bd6553ffdd/README.md).

```java
public class TestFxTest extends ApplicationTest {

    @Override
    public void start(Stage stage) {
        Label label = new Label("Hello world");
        label.setId("greeting");

        Scene scene = new Scene(label);
        stage.setScene(scene);
        stage.show();
    }

    @Test
    public void shouldDisplayHelloWorld() {
        FxAssert.verifyThat("#greeting", LabeledMatchers.hasText("Hello world"));
    }
}
```

This fail with : 
```
java.lang.AssertionError: 
Expected: Labeled has text "Hello world"
     but: was null
```

Could you explain me why ?

brcolow:
Can you post a full project (with build files, imports, etc.) that demonstrates this?

jcornaz:
Here you are : https://github.com/jcornaz/testfx-test

The code in this project is an exact copy-past from here : https://github.com/TestFX/TestFX/wiki/Version-4:-Lifecycle, and it is failling as well.

brcolow:
You must actually launch the application. Add:

```java
    @BeforeEach
    void setup() throws Exception {
        ApplicationTest.launch(ClickApplication.class);
    }
```

To `ClickApplicationTest.java`

jcornaz:
Ok. Thank you for you.

Please add it to the readme as well.

Before closing this issue, could you tell me why is the purpose of overriding the `start` method if it is needed to launch the application (which will setup the stage too) ?

jcornaz:
After testing your proposition, it doesn't fix the problem.

Now, I see the application launched, but the application is hanging, and the test are not blocked.

Note: I have update the example project, if you want to see it

brcolow:
Here is a working project:

https://github.com/brcolow/testfx-test

I made some changes. I don't know which change exactly is the key as I am quite pressed for time. Perhaps you could investigate and let me know.

Thanks.

jcornaz:
OK. I'll take a look.

Thank you.

jcornaz:
Ok. I made a mistake in my last try, I used `Application` instead of `ApplicationTest` to launch my application before each test.

In your working project example, the change that makes the difference is adding the following in the test class :

```java
    @BeforeEach
    public void setup() throws Exception {
[...]

PR #476, opened 2017-12-10
Correct usage example in readme

jcornaz:
As shown in #475, the current code example in the readme is misleading.

Overriding `start` is not enough to test the UI. It is necessary to launch the application under test with `ApplicationTest.launch()`

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,11 +36,10 @@ online (these are auto-generated from `master`) or by running `gradle javadoc`.
 
 ~~~java
 public class DesktopPaneTest extends ApplicationTest {
-    @Override
-    public void start(Stage stage) {
-        Scene scene = new Scene(new DesktopPane(), 800, 600);
-        stage.setScene(scene);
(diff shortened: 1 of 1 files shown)


brcolow:
Ideally we would want to keep the code in that set's up the scene, though. So the user doesn't have to guess how the scene is setup from DesktopPane.

jcornaz:
Indeed, using `@ExtendWith(ApplicationExtension.class)` resolves my problem.

But the readme is still misleading. It says that Junit5 is supported, but the example only works for Junit4, and there is no clue to understand that in the readme.

I propose this kind of change : https://github.com/jcornaz/TestFX/commit/9711fe4298c28d91d0c5d2c2ce75d743eb398a4d

brcolow:
Okay makes sense, can you open a new PR or adapt this one?

jcornaz:
Ok, I created #478.

PR #478, opened 2017-12-12
Add notice for Junit5 users in readme

jcornaz:
Currently, the readme says TestFx is compatible with Junit5, but the example is only valid for Junit4.

This PR add a link to a Junit5 example and a notice to use the `testfx-junit5` module with Junit5.


brcolow:
Looks good - thanks!

PR #501, opened 2018-01-02
(enhancement) Improve TableViewMatchers containsRow javadocs.

### LINKED ISSUE #499: TableViewMatchers.containsRow method behaviour
Hello there,

I've started to use your great library to test my small application and currently facing weird behaviour of the TableViewMatchers.containsRow method. As name applies, it should return true if TableView has entry which matches to the row, entered as Object... parameter.
For example, I'm expecting true when TableView contains 
```
row 1 -> "Blue", 1, false
row 2 -> "Red", 4, true
```
and I'm looking for row to have values -> "Blue", 1, false.
However method returns true, as it expects both row to be looked for and order is compulsory.
Is it how it should behave? Then the name is misleading :(

brcolow:
Fixes #499.

### DIFF
--- a/subprojects/testfx-core/src/main/java/org/testfx/matcher/control/TableViewMatchers.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/matcher/control/TableViewMatchers.java
@@ -87,8 +87,29 @@ public class TableViewMatchers {
     }
 
     /**
-     * Creates a matcher that matches all {@link TableView}s that have a row at the given {@code index} with
-     * all of the given cells.
+     * Creates a matcher that matches all {@link TableView}s that have a row at the given {@code index} that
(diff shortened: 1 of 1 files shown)


MikhailMS:
It seems like there could be a small error in description
You saying, that 

> 
verifyThat("#fitnessTable", TableViewMatchers.containsRow("Dan Anderson", 28.83, true);
verifyThat("#fitnessTable", TableViewMatchers.containsRow("Dan Anderson", "28.83", "true");
> 

will both work, but I've tried it in my app and it fails. Seems like it compares not only values (which are indeed same) but also type of the Object (which is not same) and fails. :/

brcolow:
Ah yes, that only works for hasTableCell, not sure why I made that mistake. Will correct docs. Thanks!

PR #510, opened 2018-01-18
(chore) Cleanup MoveRobotImpl motions.

brcolow:
Investigating https://github.com/TestFX/TestFX/issues/506

PR #511, opened 2018-01-20
MoveRobotImpl slowness fix

seggcsont:
I have created this PR because it was implemented in our test code as a workaround.


brcolow:
Looks good. Better than my version in #510 👍 - thanks very much!

PR #530, opened 2018-02-15
(fix) Fix the region that is used to capture Bounds.

psaikkonen:


brcolow:
Good catch! Thanks very much.

PR #543, opened 2018-03-10
Windows 10 scaling

### LINKED ISSUE #245: Windows 10 scaling issue
First of all, TestFX is a great, very useful tool but seems to struggle with Windows 10 scaling. If I ask TestFX to click on a button on my hi-res screen of 2736x1824 with a scaling of 200%, it seems to click on the button at the 100% scale so it misses the button by quite a long way. If I change my scaling to 100% and restart the computer the same tests work fine.

So it seems when looking up a button it grabs the coordinates of the button at the wrong scale.

By the way, I am using JDK 8u71 and testFX 4.0.1-alpha. Is there a simple argument I can apply to work around this?

Aidamina:
Attempting to fix #245

Not confident yet that this is the right approach.

### DIFF
--- a/subprojects/testfx-core/src/main/java/org/testfx/service/locator/impl/BoundsLocatorImpl.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/service/locator/impl/BoundsLocatorImpl.java
@@ -20,11 +20,15 @@ import javafx.geometry.BoundingBox;
 import javafx.geometry.Bounds;
 import javafx.scene.Node;
 import javafx.scene.Scene;
+import javafx.stage.Screen;
 import javafx.stage.Window;
 
 import org.testfx.api.annotation.Unstable;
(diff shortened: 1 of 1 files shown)


brcolow:
Do you think you could add a test that was failing before, but passes now? I really like to add that when possible so that we don't regress this in the future. Thanks for taking this up, it's greatly appreciated.

brcolow:
Cherry-picked your work into #544 - working on adding tests (and CI support for hidpi) and using the methods I mentioned above if we are on Java 9 and above. Otherwise (for Java 8) we will fall back to AWT, unless there is a way to get the scale on Java 8.

Aidamina:
Sounds good, was on a holiday, so couldn't deliver on the tests.

brcolow:
No problem, thanks for getting the ball rolling. I realized that to test this properly will require quite a bit of non-trivial work with the CI infrastructure, so I am tackling that in #544.

PR #544, opened 2018-03-17
Support different UI scales (hidpi)

brcolow:


brcolow:
macOS build is failing because homebrew bumped the java formula to JDK 10. Currently there is no "java9" in `caskroom/versions`: https://github.com/caskroom/homebrew-cask/pull/45013 so not sure how to explicitly install java9 via homebrew for that build.

PR #546, opened 2018-03-23
(fix) Test FxRobot.push(..) with shortcut combo.

brcolow:
From gitter:

```
Danny Gonzalez
@dannyagonzalez_twitter
Trying to execute file open on an application I’m testing with FxRobot. The menu item has an accelerator set to KeyCodeCombination(KeyCode.O, KeyCombination.SHORTCUT_DOWN). When I use FxRobot.push(new KeyCodeCombination(KeyCode.O, KeyCombination.SHORTCUT_DOWN)) this seems to work fine on Windows but not on the Mac, any reason why?
```

### DIFF
--- a/subprojects/testfx-core/src/test/java/org/testfx/cases/integration/MenuBarTest.java
+++ b/subprojects/testfx-core/src/test/java/org/testfx/cases/integration/MenuBarTest.java
@@ -51,6 +51,7 @@ public class MenuBarTest {
     FxRobot fxRobot = new FxRobot();
     Menu editMenu;
     CountDownLatch newMenuShownLatch = new CountDownLatch(1);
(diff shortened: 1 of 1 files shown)


liamsharp:
Any thoughts on when this one will be merged and available in a release?

liamsharp:
@brcolow Thanks for looking into this one. I've worked with Danny that reported this initially in gitter to create a simple maven project that illustrates this issue, you can get it here:

https://github.com/liamsharp/testfx-accelerators/

Works on Windows 10 & Ubuntu 16.04, but not on macOS 10.13.5.

If we replace `KeyCombination.SHORTCUT_DOWN` with `KeyCombination.META_DOWN` the test passes.

liamsharp:
@brcolow I've filed a bug will a full explanation in #589

PR #558, opened 2018-04-10
(chore) Add Java 10 builds to Appveyor and Travis CI.

brcolow:


brcolow:
I realized that the Windows CI builds were never actually running headlessly because of the wrong environment variable (JAVA_OPTS vs _JAVA_OPTIONS). In changing it some scale-related bugs pop up, so trying to see what the deal is there.

brcolow:
Headless builds on 9 and 10 will be broken till https://bugs.openjdk.java.net/browse/JDK-8196031 is fixed.

brcolow:
Rendering text on headless Java 10 is causing a JVM crash because `-Dprism.text=t2k` is no longer available in Java 10, so it uses native text rendering: https://github.com/javafxports/openjdk-jfx/issues/66

brcolow:
As suggested by Phillip Race on JBS, using d3d seems to work except for the capture tests. I am wondering why we ever used prism.order=sw now, do you know @hastebrot ?

PR #570, opened 2018-04-29
Test.

brcolow:


brcolow:
Even though XVFB is used to create multiple screens Java only sees one so this test is a failure. Was looking for a way to simulate multiple screens using CI.

PR #576, opened 2018-05-17
(doc) Reword doc to match behaviour.

scolej:
It seems that the doc on the `match()` methods is backwards.

The methods keep matching nodes, rather than removing matching notes.

Thanks for this library!


brcolow:
Very true. Thanks for the PR!

Issue #577, opened 2018-05-18
Check text in MenuItem elements


Alexilmarranen:
How can I check the text in MenuItem in the following configuration:
junit - 5.2.0, testfx - 4.0.13-alpha
In MainApp MenuBar contain Menu "File"(id:fileMenu), Menu "File" contain Menu "New"(id:newMenuItem).
I run this test:

```
@ExtendWith(ApplicationExtension.class)
class RootLayoutControllerTest {

    @Start
    void onStart(Stage stage) throws Exception {
        MainApp app = new MainApp();
        app.start(stage);
        stage.show();
    }

    @Test
    void menuBarHasTextInItems(FxRobot robot) {
        verifyThat("#fileMenu", org.testfx.matcher.control.LabeledMatchers.hasText("File")); //Success
        robot.clickOn("#fileMenu");//Success
        robot.moveTo("#newMenuItem", Motion.DEFAULT);//Success
        verifyThat("#newMenuItem", org.testfx.matcher.control.LabeledMatchers.hasText("New"));//Error
    }
}
```
and error: 

>java.lang.AssertionError: 
Expected: Labeled has text "New"
     but: was a com.sun.javafx.scene.control.skin.ContextMenuContent$MenuItemContainer (<ContextMenuContent$MenuItemContainer[id=newMenuItem, styleClass=menu-item]>)
Expected :Labeled has text "New"

Any ideas how check this text?
Sorry if my question is dumb

ribosradagast:
adding your code that sets up the button / menu and menu items would be very helpful

Alexilmarranen:
I do not use programmatically set up of components - I load FXML file and init scene.
However, I wrote an identical example to my code with the same error.

```
   @Start
    void onStart(Stage stage) {
        MenuBar menuBar = new MenuBar();

        Menu menu = new Menu("File");
        menu.setId("fileMenu");

        MenuItem menuItem = new MenuItem("New");
        menuItem.setId("newMenuItem");
        menu.getItems().add(menuItem);

        menuBar.getMenus().add(menu);
        stage.setScene(new Scene(new StackPane(menuBar), 100, 100));
        stage.show();
    }

    @Test
    void menuBarHasTextInItems(FxRobot robot) {
        verifyThat("#fileMenu", hasText("File"));
        robot.clickOn("#fileMenu");
        robot.moveTo("#newMenuItem", Motion.DEFAULT);
        verifyThat("#newMenuItem", hasText("New"));//Error
    }
```

[...]

PR #584, opened 2018-06-14
Add new matcher for MenuItem elements with tests

Alexilmarranen:
Partial solution of the problem in issue #577

### DIFF
--- a/subprojects/testfx-core/src/main/java/org/testfx/matcher/control/MenuItemMatchers.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/matcher/control/MenuItemMatchers.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2013-2014 SmartBear Software
+ * Copyright 2014-2018 The TestFX Contributors
+ *
+ * Licensed under the EUPL, Version 1.1 or - as soon they will be approved by the
+ * European Commission - subsequent versions of the EUPL (the "Licence"); You may
+ * not use this work except in compliance with the Licence.
+ *
+ * You may obtain a copy of the Licence at:
+ * http://ec.europa.eu/idabc/eupl.html
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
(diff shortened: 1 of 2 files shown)


brcolow:
Very cool. Do you think it would be possible for you to also add the equivalent AssertJ assertions layer? What's involved is adding an abstract class to `subprojects/testfx-core/src/main/java/org/testfx/assertions/api/` called `AbstractMenuItemAssert` and an extending class `MenuItemAssert`. These would be analogous to https://github.com/TestFX/TestFX/blob/master/subprojects/testfx-core/src/main/java/org/testfx/assertions/api/AbstractTableViewAssert.java and https://github.com/TestFX/TestFX/blob/master/subprojects/testfx-core/src/main/java/org/testfx/assertions/api/TableViewAssert.java (but for MenuItem instead of TableView, of course). Then an `assertThat` overload needs to be added to https://github.com/TestFX/TestFX/blob/master/subprojects/testfx-core/src/main/java/org/testfx/assertions/api/Assertions.java such as:

```java
/**
 * Create assertion for {@link MenuItem}.
 *
 * @param actual the actual value
 * @return the created assertion object
 */
public static MenuItemAssert assertThat(MenuItem actual) {
    return new MenuItemAssert(actual);
}
```

It's be a little bit since I did any TestFX related work and so I am not sure if there are any specific reasons why a non-Node (`MenuItem` happens to not be a `Node` descendant) can't have AssertJ assertions. It may require also adding an `AbstractEventTargetAssert` class that `AbstractMenuItemAssert` would extend from.

If that's too much to ask then we can merge this as-is but MenuItem matching will only be supported with Hamcrest-style matchers and not AssertJ assertions.

Thanks!

Alexilmarranen:
I will be glad to do this layer but it will take some time. Many thanks for the reply and the detailed explanation. I will try to do all this and I will do PR.

brcolow:
Okay, that works for me. Thanks, merged.

Alexilmarranen:
I was going to make the promised, I made fetch upstream and found out that AssertJ assertions are made by you. I apologize for not fulfilling the promise. (

Issue #590, opened 2018-07-12
Unable to paste when running headless on macOS


liamsharp:
## Expected Behavior

To be able to call:

`robotContext().getTypeRobot().push(pasteKeyCombo);`

and have it past the text in.

## Actual Behavior

it doesn't work, and ends up entering the character v from the copy and paste short cut.

Demo project here:

https://github.com/liamsharp/testfx-pasting

Running (Headfull):

    mvn clean test 

works on Windows 10 & Ubuntu 16.04 and macOS 10.13.5.

Running (Headless)

    mvn clean test -PHeadless

works on Windows 10 & Ubuntu 16.04, but fails macOS 10.13.5.

## Specifications

  - Version: 10.13.5
  - Platform: macOS

bob80333:
Have you tried using `KeyCombination.SHORTCUT_DOWN`?

From the KeyCombination [javadocs](https://docs.oracle.com/javase/8/javafx/api/javafx/scene/input/KeyCombination.html):

> The shortcut modifier is used to represent the modifier key which is used commonly in keyboard shortcuts on the host platform. This is for example control on Windows and meta (command key) on Mac. By using shortcut key modifier developers can create platform independent shortcuts. So the "Shortcut+C" key combination is handled internally as "Ctrl+C" on Windows and "Meta+C"

I modified the `getPaste()` method in your demo project to the following, and it started working correctly for me.

```
    private KeyCodeCombination getPaste()
    {
        return new KeyCodeCombination(KeyCode.V, KeyCombination.SHORTCUT_DOWN);
    }
```

bob80333:
By checking SystemUtils.JAVA_AWT_HEADLESS, there is a workaround.

```private KeyCombination.Modifier getModifier()
    {
        return SystemUtils.IS_OS_MAC ? ("true".equals(SystemUtils.JAVA_AWT_HEADLESS) ? KeyCombination.SHORTCUT_DOWN : KeyCombination.META_DOWN) : KeyCombination.CONTROL_DOWN;
    }```

liamsharp:
@bob80333 Excellent, that workaround does the trick.

brcolow:
Should be fixed with #591

Issue #436, opened 2017-10-05
Build of testfx-core on Mac OS Sierra 10.12.6 fails


Oliver-Loeffler:
When trying to build TestFX on Mac OS Sierra 10.12.6, the build will fail due to failing tests in **testfx-core** subproject.

Following tests are affected:
 * ShortcutKeyTest
```
org.testfx.robot.impl.ShortcutKeyTest
   .shortcut_keyCode_converts_to_OS_specific_keyCode_when_pressed(ShortcutKeyTest.java:107)

org.testfx.robot.impl.ShortcutKeyTest
   .shortcut_keyCode_converts_to_OS_specific_keyCode_when_released(ShortcutKeyTest.java:116)
```

 * AwtRobotAdapterTest
```
org.testfx.service.adapter.impl
   .AwtRobotAdapterTest.getCapturePixelColor(AwtRobotAdapterTest.java:260)

org.testfx.service.adapter.impl
   .AwtRobotAdapterTest.getCaptureRegion(AwtRobotAdapterTest.java:275)
```
 * GlassRobotAdapterTest
```org.testfx.service.adapter.impl.GlassRobotAdapterTest
   .getCaptureRegion(GlassRobotAdapterTest.java:274)
```

Subprojects **testfx-junit** and **testfx-junit5** are passing all checks.

Oliver-Loeffler:
See #411, #431, #435.

Oliver-Loeffler:
I have some news regarding the failing `GlassRobotAdapterTest` and `AwtRobotAdapterTest` classes. It seems that this issue on macOS 10.12.6 in conjunction with JDK9 is related to color management.

The idea is, when a component with a color "#abcdef" is drawn on screen, a capture of this colored region should still have the same color when tested. Of course, when color management is used, the value actually drawn to the screen might be different, so there are in fact somewhere at least 2 color values.

Now, I wrote a small test where the region color is modified from (e.g.)  `Color.rgb(0,0,0)` to `Color.rgb(255,255,255)` or to `Color.rgb(0,255,0)`. Lets call the desired region color value "input" and the color value of the tested pixel "captured". Now the input value vs. the captured value can be compared, per color channel (e.g. only comparing the red component). So it's now possible to plot input-red vs. captured-red and input-(green, blue) versus captured-(green,blue).

[...]

PR #591, opened 2018-07-17
589 shortcut

### LINKED ISSUE #589: Menu Accelerator using KeyCombination.SHORTCUT_DOWN doesn't work on macOS
## Expected Behaviour 

Adding an accelerator to a `MenuItem`, eg:

`openMenuItem.setAccelerator(new KeyCodeCombination(KeyCode.O, KeyCombination.SHORTCUT_DOWN));`

should be triggered by the following testFx code:

`robotContext().getTypeRobot().push(new KeyCodeCombination(KeyCode.O, KeyCombination.SHORTCUT_DOWN));`

but it's not. This works on Windows 10 & Ubuntu 16.04, but fails macOS 10.13.5

## Actual Behavior

Example project here:

https://github.com/liamsharp/testfx-accelerators/

Run:

`mvn clean test`

and you get a failure. If you change `AppTest` to use `KeyCombination.META_DOWN` and it works.

## Specifications

  - Version: 10.13.5
  - Platform: macOS

Ortner:
## Issue
Command-Key handling on mac. On my platform the tests in `ShortcutKeyTest.java` on master fail. I guess they are running on Travis somehow. There are pretty much issues relating to the mac `SHORTCUT` key (#589, #436, #590).

Per specification, the `COMMAND`-Key on Mac does map to the `META` KeyCode in Java. The CONTROL key on Windows, well to the `CONTROL` KeyCode. Both Map to the `SHORTCUT` KeyCode in Java. The `SHORTCUT` is a generated Property by the JVM (a virtual, non existing key), it may not be used as real input. 

## Behaviour of KeyCodes on macOS

- `press(KeyCode.META)` - Received by KeyEvent: `KeyCode.COMMAND`, `e.isShortShortcutDown()`: true
- `press(KeyCode.SHORTCUT)` - Received by KeyEvent: `KeyCode.A`, `e.isShortShortcutDown()`: false (That's where all those 'a's in my text editor come from after running TestFX-Tests)
- `press(KeyCode.CONTROL)` - Received by KeyEvent: `KeyCode.Ctrl`, `e.isShortShortcutDown()`: false
- `press(KeyCode.COMMAND)` - Received by KeyEvent: nothing at all

## Expected Behavior

- `SHORTCUT+A`, `SHORTCUT+C`, `SHORTCUT+V` (...) should work
- the method KeyEvent.isShortcutDown() should actually return true while pressed (currently it returns false)
- KeyCode Combinations should work (e.g. robotContext().getTypeRobot().push(new KeyCodeCombination(KeyCode.O, KeyCombination.SHORTCUT_DOWN));) (#590, #589)

## Solution
Sending `KeyCode.META` (macOS) and receiving `KeyCode.COMMAND`. Added test for the other points too.

## Related issues
fixes #589 #590  (forgot to close in commit message)
partially #436

### DIFF
--- a/subprojects/testfx-core/src/main/java/org/testfx/api/FxRobot.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/api/FxRobot.java
@@ -515,7 +515,7 @@ public class FxRobot implements FxRobotInterface {
         context.getSleepRobot().sleep(duration, timeUnit);
         return this;
     }
-
+ 
     @Override
     @Deprecated
     public FxRobot scroll(int amount) {
--- a/subprojects/testfx-core/src/main/java/org/testfx/robot/impl/KeyboardRobotImpl.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/robot/impl/KeyboardRobotImpl.java
@@ -31,7 +31,15 @@ import org.testfx.util.WaitForAsyncUtils;
 
 public class KeyboardRobotImpl implements KeyboardRobot {
 
-    protected static final KeyCode OS_SPECIFIC_SHORTCUT = System.getProperty("os.name").toLowerCase(Locale.US)
+    /**
+    * This key is sent depending on the platform via the Robot to Java.
+    */
+    private static final KeyCode OS_SPECIFIC_SHORTCUT = System.getProperty("os.name").toLowerCase(Locale.US)
+            .startsWith("mac") ? KeyCode.META : KeyCode.CONTROL;
+    /**
+    * This key is received by key listeners as os specific shortcut key.
+    */
+    protected static final KeyCode OS_SHORTCUT = System.getProperty("os.name").toLowerCase(Locale.US)
             .startsWith("mac") ? KeyCode.COMMAND : KeyCode.CONTROL;
 
     private final BaseRobot baseRobot;
--- a/subprojects/testfx-core/src/main/java/org/testfx/robot/impl/KeyboardRobotImpl.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/robot/impl/KeyboardRobotImpl.java
@@ -36,11 +36,6 @@ public class KeyboardRobotImpl implements KeyboardRobot {
     */
     private static final KeyCode OS_SPECIFIC_SHORTCUT = System.getProperty("os.name").toLowerCase(Locale.US)
             .startsWith("mac") ? KeyCode.META : KeyCode.CONTROL;
-    /**
-    * This key is received by key listeners as os specific shortcut key.
-    */
-    protected static final KeyCode OS_SHORTCUT = System.getProperty("os.name").toLowerCase(Locale.US)
-            .startsWith("mac") ? KeyCode.COMMAND : KeyCode.CONTROL;
 
     private final BaseRobot baseRobot;
     private final Set<KeyCode> pressedKeys = ConcurrentHashMap.newKeySet();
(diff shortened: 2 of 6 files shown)


brcolow:
Thanks for stepping up on this one, @Ortner . I note two things that I'm sure you are aware of:

1.) Currently the PR can't be merged because it is on a stale head. Can you rebase to latest master?
2.) The shortcut tests fail on Travis on macOS. If this is some sort of weird thing that only happens on Travis we can ignore the test for Travis specifically, but it would be nice to know the reason why.

Ortner:
Hi,

1.) I really didn't notice the stale branch. Sorry for that, Might be better now.

2.) I hate those issues. By the way, as I don't have access to the files, where the test output is stored (do I?), is it possible to keep the output in the console of travis if the build failes?

## Solution

It seems, that the reported key depends on the setup of the Mac (OS/Java version). On My System, the `COMMAND`-Key is reported, Travis gets the `META`. As I need only the code to verify, that actually one of the Shortcut keys is pressed, when evaluating the `isShortcutDown`-method, I modified the code, to accept all possible `KeyCode` values.

Ortner:
Seems to be an issue of the used Robot. My system uses the `AWTRobot` and Travis uses the `GlassRobot`. 

When switching to `GlassRobot`, the test fails with `META` and succeeds with `COMMAND` on my system too.

Ortner:
I would suggest the following convention:
 - KeyBoardRobot: handles real keys and virtual/modifier keys as input and converts them to (platform specific real keys (that exist on the keyboard). This ensures a common interface to the robot and the user may request virtual keys e.g. shortcut or meta)
 - RobotAdapter: handles real keys (platform dependent, existing on a keybord) as input. It might modify the key as requiered by the robot

Ortner:
Can you please retrigge the AppVeyor test.
```
# Problematic frame:
# C  [javafx_font.dll+0x4e80]
```

SigSegv seems to be unrelated to this code change I assume.

Ortner:
@brcolow : is it possible to run the travis build for mac with the AwtRobot, to catch those issues?
Currently on my System some more tests are failing:
 - the color tests (#436, I'm going to have a look at that too)
 - and the Font tests

So expect them to fail in case of AWTRobot

brcolow:
Experimenting with a headless Mac build using awt robot: https://github.com/TestFX/TestFX/pull/596
(should default to "awt" when "testfx.robot" property is empty).

brcolow:
When using AWT Robot on Travis, because it is headless, we trigger a `HeadlessException` everytime `getMouseLocation` is called. Also we explicitly throw a `RuntimeException` when trying to create an AWT robot when running heedlessly:

```
    @Override
    public void robotCreate() {
        if (GraphicsEnvironment.getLocalGraphicsEnvironment().isHeadlessInstance()) {
            throw new RuntimeException("can not create awt robot as environment is headless");
        }
        Toolkit.getDefaultToolkit(); // initializes awt toolkit
        awtRobot = createAwtRobot();
    }
```

Should we get rid of the exception (maybe replace it with a warning) and have the `getMouseLocation` method return `(-1, -1)` if running in headless mode? Or should we leave it as is? Unfortunately it is not possible to run in headed mode on macOS on Travis CI.

brcolow:
Good job on fixing the failing tests. The branch still shows to be in conflict.

It should be possible to fix this via:

```sh
git remote add upstream https://github.com/testfx/testfx
git fetch upstream
git rebase upstream/master
# fix conflicts
git push origin --force
```

Ortner:
I needed to add some changes t the files, as rebase --continue didn't work without changes. Does it work now?
If not I'm going to create a new branch...

brcolow:
No conflicts now.

brcolow:
Merged. Thanks very much @Ortner

PR #594, opened 2018-07-17
(fix) typing of characters in the background

### LINKED ISSUE #593: Modification of text editor when running tests in IDE
## Expected Behavior

Do not modify my text editor in the background when running the test suite

## Actual Behavior

Always have an annoying 'a' in the source code after running the standard test suite.

Ortner:
## Issue

Text is typed in my editor when running the testFX testsuite (see #593).

## Solution

Added a component in the background for the basic tests.

## Open issue

I needed to add (and use) a java package, as checkstyle wants to have a line between javafx imports and the (not existing) group before it.

fixes #593

### DIFF
--- a/subprojects/testfx-core/src/test/java/org/testfx/cases/TestCaseBase.java
+++ b/subprojects/testfx-core/src/test/java/org/testfx/cases/TestCaseBase.java
@@ -16,15 +16,42 @@
  */
 package org.testfx.cases;
 
+import java.util.List; //checkstyle needs one line between javafx and any group before it
+
+import javafx.geometry.Insets;
+import javafx.scene.Scene;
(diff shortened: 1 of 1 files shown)


brcolow:
That Checkstyle bit is annoying, we can use an in-line comment block to suppress the warnings:

```
//CHECKSTYLE:OFF
import blah
//CHECKSTYLE:ON
```

We could also maybe tweak the `checkstyle.xml` ruleset to allow for this import situation - but I am not sure how.

brcolow:
Merged in https://github.com/TestFX/TestFX/pull/598 (just removed the Checkstyle bit as it appears forcing the java import wasn't needed after all). Authorship should be retained.

PR #606, opened 2018-08-06
[enhancement] Support Java(FX) 11.

brcolow:


brcolow:
Now that the build is working with the Maven artifacts, the next step is to add the necessary stuff to `JavaVersionAdapter` to seamlessly support the old private API robot and the new public API robot, which I will do soon.

brcolow:
Next I need to issue a new monocle release for Java 11, then headless builds should pass.

brcolow:
Hmmm looks like I may have broke screen capture on Monocle in https://github.com/javafxports/openjdk-jfx/pull/36 - 

```
Caused by: java.lang.ArrayIndexOutOfBoundsException: Index 20480 out of bounds for length 20000
    	at com.sun.glass.ui.monocle.MonocleRobot.getScreenCapture(MonocleRobot.java:217)
    	at com.sun.glass.ui.GlassRobot.getScreenCapture(GlassRobot.java:194)
    	at javafx.scene.robot.Robot.getScreenCapture(Robot.java:302)
```

I'll look into that.

brcolow:
PR opened upstream: https://github.com/javafxports/openjdk-jfx/pull/155

subprojects/testfx-core/testfx-core.gradle
```diff
@@ -16,20 +16,46 @@
...
+ext.openjfxVersion = '11-ea+19'
 
 repositories {
     jcenter()
+    maven { url 'https://jitpack.io' }
 }
 
+static def getOSName() {
+    final String osName = System.getProperty("os.name").toLowerCase()
+    if (osName.contains("linux")) {
```
mwkroening: Maybe it would be cleaner to use gradle's OperatingSystem interface like this:
https://stackoverflow.com/a/31443955

mwkroening: Ah I see! This is how it's done in https://github.com/johanvos/jfx-samples .
Nevermind then. I suspect, it is because it's incubating. I will have another take on this in https://github.com/johanvos/jfx-samples once it's stable.

brcolow:
Just waiting for the fix to be pulled into upstream (and ideally a new release tag to be made so I don't have to use jitpack to fetch a commit hash) and then I will release a new version of TestFX's re-packaged Monocle that has this fix, then this PR can proceed.

brcolow:
Seems like the OpenJDK 11 headless builds stall after starting `testfx-junit` test suite (no output gets received for 10 minutes, so Travis fails the build). Will try and add some debugging to drill down into that.

brcolow:
Well, trying to call jstack on Travis has turned into quite the task 😆. From my hacky timer thread dumping code it seems like, on the Java 11 headless macOS build, the deadlock on the main JavaFX thread is on a CountDownLatch. But because the stack trace is cut-off, I am not sure. I will try and find a way to get the full stack traces without using jstack.

brcolow:
Alright got the green light so that's good. However I am not at all happy with the sloppy way I implemented this. It was more just a proof of concept to get the green light and see what it would take to support Java 11. Now I will focus on making the approach more clean. The first thought I have is to have two separate classes, `PublicGlassRobotAdapter` and `PrivateGlassRobotAdapter`, but not sure about that.

brcolow:
@Ortner When you are back from holidays (I know I am probably overloading you at this point - sorry about that!) would you mind taking a quick look at this when you are back (just as a sanity check)? Thanks.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.7-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.10-bin.zip
```
mwkroening: Note that using the [bugfix release 4.10.2](https://github.com/gradle/gradle/releases/tag/v4.10.2) over 4.10.1 or 4.10 is being recommended by the gradle team.

kleopatra:
Looking at the code in GlassRobotAdapter, it'll blow on destroy with the public fx Robot (that's what I meant over in the other issue - wasn't too clear, though ;)

```
    @Override
    public final void robotDestroy() {
        if (glassRobot != null) {
            waitForAsyncFx(RETRIEVAL_TIMEOUT_IN_MILLIS, () -> {
                try {
                    getRobot().getClass().getMethod("destroy").invoke(glassRobot);
                }
                catch (IllegalAccessException | InvocationTargetException | NoSuchMethodException e) {
                    throw new RuntimeException(e);
                }
                glassRobot = null;
            });
        }
    }

```
for PublicRobotAdapter, the referenced Robot has no method _destroy_ - to destroy the underlying "real" glassRobot, the code needs to retrieve the value of Robot's _peer_ field and destroy that.

Wondering why you don't leave the destroy to the concrete implementation, just the same as the create?

brcolow:
Indeed you are correct that this will break for the public robot but:

1.) Nothing currently calls robotDestroy and it's not possible to get an instance of RobotAdapter using the TestFX API (without reflection).

2.) Using reflection to fetch the peer and then destroy it is kind of ugly as it sort of bypasses the public API nature. That's why I mentioned in the other issue that we are investigating adding a disposer mechanism in [JDK-8207373](https://bugs.openjdk.java.net/browse/JDK-8207373) so that the robot will clean up itself.

I'm not saying that the issue isn't worth fixing, it definitely is and thank you for pointing it out - I just wanted to say that it shouldn't create any problems in the immediate future.

brcolow:
@kleopatra Opened https://github.com/TestFX/TestFX/pull/627 to fix this issue. Thank you.

Ortner:
@brcolow 
The deadlock issue: This should be most propably the dealock issue I mentioned/fixed in #610. The AsyncUtils use a CountDownLatch to lock the JavaFx Thread. If this method is called from the FX-Thread, it will result in a deadlock. How did you fix it, as there is only one big commit left, it's quite hard to find?

brcolow:
I believe the deadlock issue I was seeing was a red herring that was being caused by not having the module path setup correctly to contain the JavaFX modules from Maven Central. Once I had the OpenJFX dependency setup correctly for Java 11 the "deadlock" went away. The only remaining deadlock that I know about is of a different sort and it is what prevents using the headed glass robot on Appveyor (not on local Windows machine).

brcolow:
In `appveyor.yml` in the root directory of this repository you will see a bunch of commented-out builds in the matrix, for example:

```yaml
    # Java 10 / Glass Robot (deadlock?)
    # - JAVA_VERSION: "10"
    #  JAVA_HOME: C:\jdk10
    #  _JAVA_OPTIONS: "-Dtestfx.robot=glass"
```

Just remove the comment hashes from whichever Java version you want to test and have Appveyor build the PR as usual.

Issue #607, opened 2018-08-07
Clean Handling of HDPI


Ortner:
## Issue
I'm currently migrating to Java 10 and some tests on mac are failing. One of them is the related to HDPI. At first, thanks @brcolow to implement and test the HDPI feature lastly, but the solution has one drawback: The GlassRobotTest fails on my system (that uses AWTRobot), as the wrong robot is detected. In the BoundsLocator routine.

I would propose a change in the architecture based on the following requirements/expectations

## Requirements
- The user of TestFx is generally expecting the coordinate space to be within JavaFx coordinates (regardless of the Monitor being HDPI or not).
- Common Interface for the RobotAdapter - In my Opinion, the RobotAdapter interface should be consistent, so all RobotAdapters should use the same coordinate space. (On first inspection, JavaFxRobot and AwtRobot are using  JavaFx coordinates and GlassRobot is using real screen coordinates in some cases)

## Proposal

Use JavaFx Coordinates down to the RobotAdapter. The RobotAdapter should handle the conversion depending on Java version and operating system.

Points that may be confusing about this:
- getBoundsOnScreen returns JavaFx coordinates

Advantages:
- general code is independent of used Adaptor

Disadvanges
- HDPI Screenshots wouldn't be possible (may be add a Raw option to the RobotsCaptureRegion?)

PR-Follows

brcolow:
@Ortner Just wanted to report that HiDPI contract test has one failure on Windows with 200% DPI scaling (using AWT robot) on OpenJFX 11: https://ci.appveyor.com/project/testfx/testfx/build/master%201487/job/a9pnngw48kb9ksh3#L1051 - would you care to take a look and see if it's an easy fix? I ask because you are the most familiar with when/why scaling is needed. Thanks a lot.

Ortner:
Gonna have a look at it when i‘m back from holydays...

brcolow:
Sounds good. No rush whatsoever. Thanks.

Ortner:
@brcolow Sorry, took a while.  
First: I kept the methods above, so the scaling methods are called in the right place in the logic of the other functions in case any configuration needs it in the future (it's quite easy to mess something up there).

Second: Is there still a problem? The test seems to pass in the current build. After a quick look into the code, the dimensions should be correct, if the capturePixelColor passes (which did in the referenced build).

PR #608, opened 2018-08-07
Refactoring HDPI

Ortner:
This is the proposed solution to #607 

It should actually have the same functionality as before, but it's tested only using a Mac.

It seems, that the glass Robot on Windows uses Real (HDPI) coordinates. That would actually be in conflict with the definition of ` com.sun.glass.ui.Robot.getScreenCapture(int x, int y, int width, int height, boolean isHiDPI)`. Let's see what Travis says.

### DIFF
--- a/subprojects/testfx-core/src/main/java/org/testfx/internal/JavaVersionAdapter.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/internal/JavaVersionAdapter.java
@@ -181,7 +181,7 @@ public final class JavaVersionAdapter {
      * See the License for the specific language governing permissions and
      * limitations under the License.
      */
-    private static JavaVersion currentVersion() {
+    public static JavaVersion currentVersion() {
         if (currentJavaVersion == null) {
             currentJavaVersion = toVersion(System.getProperty("java.version"));
         }
@@ -228,7 +228,7 @@ public final class JavaVersionAdapter {
      * Before 9: http://www.oracle.com/technetwork/java/javase/versioning-naming-139433.html
      * 9+: http://openjdk.java.net/jeps/223
      */
-    private enum JavaVersion {
+    public enum JavaVersion {
         VERSION_1_1, VERSION_1_2, VERSION_1_3, VERSION_1_4,
         VERSION_1_5, VERSION_1_6, VERSION_1_7, VERSION_1_8,
         VERSION_1_9, VERSION_1_10, VERSION_11, VERSION_HIGHER;
--- a/subprojects/testfx-core/src/main/java/org/testfx/internal/PlatformAdapter.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/internal/PlatformAdapter.java
@@ -1,0 +1,39 @@
+package org.testfx.internal;
+
+import java.util.Locale;
+
+/**
+ * Provides an API for platform specific features
+ * 
+ */
+public class PlatformAdapter {
+
+    /**
+     * Stores the operating system we are running on. Shouldn't change during
+     * execution, so singleton.
+     */
+    private static OS os;
+
+    /**
+     * Gets the operting system
+     * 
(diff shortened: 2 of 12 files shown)


brcolow:
If you can get the tests to pass on all 3 platforms, then I will be happy to merge this as it's a much better approach than my hacking. When I was working on this it was somewhat of a nightmare to get all 3 platforms to behave reasonably.

Ortner:
I can see your point :-).

Well, if it's not running on the first try, I'll try to do it right. Might take some buils on the server, as failures are in the environments I don't have access to.

Ortner:
## ScratchPad Run1

### Linux

Travis died, so no logs -> next run

### Windows
Step one: Check BoundLocator(JavaOnly) and Robot Coordinates  
Unscaled Windows runs  

Java1.8 200%:
AwtRobot:
- nullOffset: 
  - JavaBounds OK
  - RobotCoordinates (click) NOK
- clickOffset: click was `1/scale`, so Robot need a factor of `scale`

GlassRobot:
- nullOffset: 
  - JavaBounds OK
  - RobotCoordinates (click) NOK
- clickOffset: click was `scale`, so Robot need a factor of `1/scale` (currently scale is applied, effectively removes this requirement)

Next step: Fix Robots and verify Window translation scaling (`defaultOffsetTest`)

Ortner:
## Intermediate Note

Some of our tests aren't setup correctly (at least for MacOS). This might be the reason for the flakyness of some Tests. 
I had to fix the BaseTest, to clean up properly (as the stages of previous test were hiding my stage). After that, the NodeMatcher didn't run any more. Reason was, the Cleanup in the test before did remove the stage. NodeMatcher did only extend FxRobot adapter and called: FxToolkit.registerPrimaryStage(); + FxToolkit.setupFixture. No stage was created to show the component in.
Also: If you change the coordinates of the Stage (`stage.setX(0);stage.setY(0);`) most of the following tests are translated too. The isolation between the tests is not very good currently.

Ortner:
## ScratchPad Run 2

### Windows

Window offset is OK  
But there is a bug in capturePixelColor for AWTRobot, as CaptureRegion works should be a Bug in pixel retrieval.


### Linux

Running Tests (excerpt)
.4 Glass with scaling Java 8 Running
.7 Java 10 noScale works

Failing Tests:
Build 9+10 are failing both JDK 10 (-Djdk.gtk.version=3?) one uses glass other one awt, so one log should be enough

As far as I can see, those are the only scaled Java10 tests

AwtRobot:
- nullOffset:
  - JavaBounds OK
  - RobotCoordinates (click) NOK
- clickOffset: click was 1/scale, so Robot need a factor of scale

GlassRobot:

- nullOffset:
  - JavaBounds OK
  - RobotCoordinates (click) NOK
- clickOffset: click was scale, so Robot need a factor of 1/scale 

Pretty much same as for windows. Didn't check the Linux builds for contra indication (just too many, with way too much output). Let Travis do the work.

brcolow:
Amazing work. Do you mind squashing to one commit? Thanks @Ortner

brcolow:
I went ahead and squashed it for you in #609 - I don't mean to insert my joint authorship on your work like that, not sure how I can have git not do that. Anyways, if it's okay with you I will merge #609 - if not and you want to squash this one, that's fine too.

Ortner:
@brcolow: Thanks for squashing, that reduces the risk, that I mess up my local branches again :-).
And I guess for your commitment to the project you deserve the co authorship anyway (by the way thanks for that). Especially in this case, you did set-up all the build servers and the first implementation this PR is based on.
So feel free to squash my PRs whenever you want :-).

brcolow:
Thanks very much. Also just to note on command-line git you are listed as the sole author, and I am just the "commiter" - so that's better IMO. Merged in https://github.com/TestFX/TestFX/pull/609

PR #613, opened 2018-08-10
(fix) KeyCombination on Mac with Java10 and AwtRobotAdapter

Ortner:
KeyCombinations for CopyPaste were not correctly recognized in the given
setup


brcolow:
Looks good, thanks a lot.

Issue #449, opened 2017-11-05
When a cell's edit mode is enabled FxRobot hangs


chinzen:
When a cell's edit mode is enabled the robot seems to start pressing some almost random keys... Like Backspace, primary mouse button, ESC, Enter, etc. 

I was able to make it stop only by manually calling the `release(new KeyCode[0])` which releases all the keys.

I have created a [repo](https://github.com/chinzen/testfx-test) with a minimal example in tornadofx/kotlin. Sorry, to lazy to rewrite it in java... But tornadofx here just call the same api, so I don't really expect a big difference. Yet yes, I can be wrong about it.

My app:
```kotlin
class MyApp : App(MyView::class)

class MyView : View() {
    private val persons = FXCollections.observableArrayList(
        Person(name = "Samantha Stuart", birthday = LocalDate.of(1981,12,4)),
        Person(name = "Tom Marks", birthday = LocalDate.of(2001,1,23))
    )
    
    override val root = tableview(persons) {
        column("name", Person::name).makeEditable()
        column("birthday", Person::birthday)
    }
}

data class Person(val name: String, val birthday: LocalDate)
```
The test cases:
```kotlin
@Test
fun `FxRobot hangs on an activation of the edit mode`() {
    val cellContent = "Samantha Stuart"
    
    // For me `doubleClickOn` breaks it and was able to make it work
    // only by calling `clickOn` twice instead.
    clickOn(cellContent)
    clickOn(cellContent)
    press(KeyCode.BACK_SPACE)
    val nameNew = "name new that is quite long"
    write(nameNew)
    press(KeyCode.ENTER)
    
    assertTrue(lookup(nameNew).tryQuery<Node>().isPresent)
}

@Test
fun `FxRobot works if the keys were manually released`() {
    val cellContent = "Samantha Stuart"
    
    clickOn(cellContent)
    clickOn(cellContent)
    releaseAllKeys()
    press(KeyCode.BACK_SPACE)
    releaseAllKeys()
    val nameNew = "name new that is quite long"
    write(nameNew)
    press(KeyCode.ENTER)
    releaseAllKeys()
    
    assertTrue(lookup(nameNew).tryQuery<Node>().isPresent)
}
```

brcolow:
I can reproduce the error. I will try and investigate it in more detail soon.

chinzen:
[...]

PR #284, opened 2016-05-25
Add WindowMatchers class

aalmiray:
Needs tests but code has been tested in production already.

Fixes #246

PR #615, opened 2018-08-13
610 implement base test case

Ortner:
Well the PR is a bit bigger than actually planned, and it will involve some discussion. There are some open issues as well.

## TestFx Core
The new base classes for tests can be found in `org.testfx.testcase.api`. Those base classes are independent of the test framework used. I did implement them in the JUnit4/5 packages for those frameworks too. Seperate classes for Components, Stages and Applications are provided. The static initializers are only defined in the description of `TestCase`, they need to be called in the frame work.

The classes for the tests of the`testfx-core` module can be found in /test/ `org.testfx.cases` as before. The required JUnit4 implementation as well as the Standard test classes for `testfx-core` are there also. 

I tried to prevent the circular dependency between project, that's why there is the extra adaption to JUnit4 within `testfx-core/test` (just for one of the test cases). I did also move the `TestFxRule` into this place (open for discussion).

## JUnit4/5 (Discussion for this PR)

I added the above mentioned tests for Components (anything extending Node), Stages and Applications and marked all others as deprecated.

I've never been a fan of the ApplicationTest, you can not extend Applications (as it already extends FxRobot), but you need to implement the Application methods (`start`, `init` and `stop`) and you need to setup everything using the `FxToolkit` yourself which might be prone to errors (see next chapter). I mostly do component test for UI components.

I think the propsed Tests are more convinient for the end user, less error prone and I guess most end users (like me) already implemented classes like this by themselfes.

I do not use the Rule Based test approach and marked it as deprecated also. If it is not widely used or if it is not a useful extension to the proposed classes, I would keep it that way and remove it at some point. Otherwise I would try to add proper cleanup (also in framework indepentent way) to those tests. 

## FxToolkit (Discussion - not for this PR)

`FxToolkit` is a great class with plenty of options to setup applications, stages, scenes, fixtures, sceneRoots cleaning up Applications, cleaning up stages and all kind of stage handling (register, show, hide etc.). But with all of those options the developer must respect the different behaviour of those methods (some show the stage some don't, reuse of stages with given coordinates and size etc.) So if not handled correctly (and even some of the tests in testfx-core didn't handle it correctly), tests become "flaky" and so does the use of `FxToolkit` (yes, I had a great time with `FxToolkit` during this PR).

First: I think there are too many options basically it would be enough to provide a method to register a stage (every UI test has a stage) and to setup a application (call the lifecycle methods, provide arguments etc.). Components or other variants can be handled by `org.testfx.testcase.api`.

Second: When some tests in `testfx-core` do not handle it correctly, shall we expect the end user to handle it correctly? It's public API. The package `org.testfx.testcase.api` handles all required interfaces to the `FxToolkit`. Each test framework should be able to adapt to those test cases. Shouldn't we use `FxToolkit` only as internal API?

## Enabled "flaky" test

I also did enable some failing/flaky tests (e.g. #284 #449), that did run on my system now and were ignored before. Also all test do only run once now, without retry.

## Open issues

- Spock: I don't have a clue about groovy, so I didn't update anything there yet and would appreciate if someone with groovy skills could do that
- Junit4/5: The Rule Option also hasn't been updated yet (needs to be done if not deprecated)

But first things first, let's see what travis says about flaky tests.


Ortner:
Guess Travis doesn't like the PR much. You should do the release without the PR first. I'll have a look at it but it might take some time.

Ortner:
Had some time. I walked through the results of one of the failing machines.

There is a possible dead lock condition in `WaitForAsyncUtils.waitForFxEvents` when called from the Fx-Application Thread. 
- First of all waiting for FxEvents on the Fx Thread is just not possible. So I fixed those methods.
- Second, dead locks are bad, so WaitForAsyncUtils now throws an Exception when trying to wait for FxEvents on the Fx Thread.

Additionally I modified WaitForAsyncUtils, that the dead lock is not possible anymore. The method also accounts for the defined time to wait, before it will enque another event (Tests run 40 seconds longer). The time may also be set to zero, which will result in the same behaviour as before.

Ortner:
## Note
- [ ] enabled previosly skipped test containAfterEdit..: still not running on travis linux machines AppVeyor seems OK
- [ ] AppVeyor: StageTestAsParent Button not found (Stage showing?)

Other mainly Timeouts (1 second is already quite high):
- [ ] CaptureSupport: Capture runs on Fx-Thread and takes quite long...
- [ ] Check valid values

subprojects/testfx-core/src/main/java/org/testfx/util/WaitForAsyncUtils.java
```diff
@@ -367,11 +376,14 @@ public static void waitForFxEvents() {
...
      * @param attemptsCount the number of attempts to try
      */
-    public static void waitForFxEvents(int attemptsCount) {
-        for (int attempt = 0; attempt < attemptsCount; attempt++) {
-            blockFxThreadWithSemaphore();
-            sleep(SEMAPHORE_SLEEP_IN_MILLIS, MILLISECONDS);
+    public static void waitForFxEvents(int attemptsCount, int pulses) {
+        if (Platform.isFxApplicationThread()) {
+            throw new RuntimeException("Waiting for FxEvents on the Fx-Thread is just not possible. " +
+                    "Call the root on the test Thread!");
```
brcolow: What is "root" referring to here? Can we make it a bit more clear? Just in case anyone encounters this error in trying to use TestFX, I like to be very explicit (erring on the side of pedantic).

Ortner: That's not pedantic, this message is meant to help the user (and us) to find the issue as fast as possible. So it shouldn't be confusing.  
The root is the method call, that actuallty triggered `waitForFxEvents`. The point is, that the method might be anywhere in the call tree:
- direct call of the user in the test case (e.g. via `interact(()->moveTo(...))` or `Platform.runLater('any RobotMethod')` the `Trace of caller`-part of the exception should point to the line
- indirect call in the TestFx framework (-> bug)
- direct call in the TestFx framework itself (-> bug)
The latter two cases might be a bit tricky to identify, but shouldn't occur that frequently (if we have done it right :-) ).

```java
throw new RuntimeException("Waiting for FxEvents on the Fx-Thread is just not possible. " +
                    "See stacktrace below for the caller (most probably 'Trace of caller')." +
                    "Try calling this method on the test thread (or any background thread)!");
``` 
This message would point to and user error first. Any better suggestions? (There are still some check-ins left, so take your time :-) )

brcolow: My wording would probably be `Waiting for events on the "JavaFX Application Thread" is not possible (nor advisable). Instead, call "waitForFxEvents" on a test thread (or any other background thread). See stacktrace(s) below to find the bad call to "waitForFxEvents".`

But yours is totally acceptable :).

Ortner:
Sure, I can create some examples on the wiki. And I think at the moment writing docu would be more fun than writing code (the linux machines on travis drive me crazy :-) ). But first I'll get the PR running - there are still a lot of open points/flaky tests. And for now I want to have all tests running without any retry. Enough for today.

brcolow:
@Ortner Let me know when you think the PR is complete and I will do a thorough review. No rush, of course.

Ortner:
I will, but it will really take some time. Currently I can't spend time on it, but I'll come back to you, when I think it's ready for discussion.

brcolow:
I think I'm seeing "Timeout in tear down method (only with awt headless): FxThread is stuck in waiting for rendering to complete. FxEvents are not processed any more." in #606 on Java 11 only though. In this case though it is not a timeout but actually seems to be a deadlock. testfx-junit seems to get stuck forever in the `@After` annotated `cleanup` method of `ApplicationLaunchTest.java`.

Ortner:
@brcolow 
> I think I'm seeing "Timeout in tear down method (only with awt headless): FxThread is stuck in waiting for rendering to complete. FxEvents are not processed any more." in #606 on Java 11 only though. In this case though it is not a timeout but actually seems to be a deadlock. testfx-junit seems to get stuck forever in the @After annotated cleanup method of ApplicationLaunchTest.java.

Yes, I adressed this problem already in ``org.testFx.util.FxRenderCondition`` for the other setups in awt headless. If you request a new pulse in AWTHeadless with no stage showing, the FxThread gets killed. In for the other setups I do not request a pulse, if the ``stage.isShowing()`` method of all stages returns false. So this seems to have changed in Java 11. I'll add a TODO to the list

There are still some other issues with the headless mode (e.g. when running all tests, sometimes the stage::onShow listener is not called at all, but the stage is showing...)

PR #616, opened 2018-08-15
(test) Remove t2k font engine usage from Travis builds.

brcolow:
This is mostly just a test to see if using the native font engines (freetype on Linux, coretext on macOS) in headless mode causes any crash, like it does on Windows. The `-Dprism.text=t2k` argument has been suggested to be used when testing in headless mode since forever, but that may just have been to have all platforms work, when it was really only needed on Windows. I am working on an [upstream fix](https://github.com/javafxports/openjdk-jfx/issues/66) to the crash that happens on Windows in headless mode (using Monocle) without `t2k` (which was removed from JavaFX in version 10, and was never open sourced, and so only ever worked with Oracle JDK and not OpenJDK).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,13 +53,13 @@ jobs:
       dist: trusty
       jdk: oraclejdk8
       env:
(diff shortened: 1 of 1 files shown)


brcolow:
Good to know this works as intended. Guess the `prism.text=t2k` recommendation was only for Windows.

PR #620, opened 2018-08-17
(fix) Remove call to limit(-1) in MoveRobotImpl.moveMouseStepwiseBetween

### LINKED ISSUE #618: stream().limit(-1) in MoveRobotImpl
The following error is produced in some cases on our CI environment when trying to click on a button:

```
java.lang.IllegalArgumentException: -1
	at java.util.stream.ReferencePipeline.limit(ReferencePipeline.java:400)
	at org.testfx.robot.impl.MoveRobotImpl.moveMouseStepwiseBetween(MoveRobotImpl.java:124)
	at org.testfx.robot.impl.MoveRobotImpl.moveTo(MoveRobotImpl.java:63)
	at org.testfx.robot.impl.ClickRobotImpl.clickOn(ClickRobotImpl.java:57)
	at org.testfx.api.FxRobot.clickOn(FxRobot.java:595)
	at org.testfx.api.FxRobot.clickOn(FxRobot.java:643)
	at org.testfx.api.FxRobot.clickOn(FxRobot.java:60)
	at org.testfx.api.FxRobotInterface.clickOn(FxRobotInterface.java:641)
	at <our test class>:<the line where it calls clickOn("#someElement")
```

I am sorry to report that I have not found out the exact reason why and under what conditions this happens, but there does not seem to be any direct indicator of what is wrong due to the lack of a proper error message.

It seems like moveMouseStepwiseBetween calls the `limit` stream method with an argument `-1`, which is caused by the `path` list being empty. I have found that "should" not be possible, since it is filled by the `interpolatePointsBetween`, which should add exactly `totalStepsCount` points to that list, which is bounded to be >= 1. Perhaps the `limit` call can be avoided completely by using a simple indexed for loop in its place.

## Expected Behavior

A proper error message is produced, that shows what caused the mouse movement issue. E.g.: the node has a negative screen position, does not exist, etc.

## Actual Behavior

See above for stack trace.

## Specifications

  - Version: 4.0.13-alpha
  - Platform: macosx10.11
  - JUnit: 4.12
  - Gradle: 4.9

Clashsoft:
The new algorithm now uses an indexed for loop instead of a stream + for
each approach. This does not have the above problem.

Fixes #618

### DIFF
--- a/subprojects/testfx-core/src/main/java/org/testfx/robot/impl/MoveRobotImpl.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/robot/impl/MoveRobotImpl.java
@@ -133,8 +133,9 @@ public class MoveRobotImpl implements MoveRobot {
             }
         }
 
-        for (Point2D point : path.stream().limit(path.size() - 1).collect(Collectors.toList())) {
-            // TODO(mike): Why is the limiting necessary?
+        for (int i = 0; i < path.size() - 1; i++) {
+            // using path.size() - 1 because the last element is always equal to the targetPoint
+            Point2D point = path.get(i);
             mouseRobot.moveNoWait(point);
             sleepRobot.sleep(SLEEP_AFTER_MOVEMENT_STEP_IN_MILLIS);
         }


brcolow:
Thank you very much!

PR #621, opened 2018-08-29
(fix) Auto-deploy via Travis - bintray fix.

brcolow:


brcolow:
The gradle-bintray-plugin was changed in a recent version to not fail silently if the root project upload fails (which it does in our case because we don't upload the root project, only subprojects, which have the bintray config). The solution in this PR is hacky (swallow the non-zero exit code so the script continues executing) and a better solution would be telling gradle/the bintray plugin to skip executing `bintrayUpload` for the root project - but I am not sure how to do that. The docs for multi-project gradle builds are sparse, and there are many historic issues on the gradle-bintray-plugin issue tracker which seem to be at least tangentially related to this issue, but it is hard to invest more time in a proper solution when the hacky one in this PR works.

brcolow:
Came up with a more verbose, but less hacky solution. Just call `bintrayUpload` on each subproject individually.

Issue #581, opened 2018-06-03
Sample TestFx not working


integrys-net:
I'm trying TestFX by run the sample test in documentation, but this not work for second test case. Button on scene isn't clicked. Mouse click on coords(0,0), that is outside of test application window.

## Specifications

  - Version: TestFX 4.0.13-alpha with JUnit 5 binding
  - Platform: Java 8

brcolow:
What operating system are you running this on, and exactly what Java version (run `java -version` to get the full string).

integrys-net:
Thank you for reply @brcolow. Th eOS is Windows 10 build 1803, and java version used to run demo test, is JDK 1.8u172.

brcolow:
Do you know the DPI of your monitor? This could be https://bugs.openjdk.java.net/browse/JDK-8196031 if you are using hiDPI.

integrys-net:
Thank you for reply, i don't know that bug. I've readed now for the first time. My screen is HiDPI. The current dpi settings is 250%.

kleopatra:
the bug seems fixed in fx11 (the test case in the bug report passes, replacing the old robot api with the current) - the simple testfx example still fails

brcolow:
@kleopatra Which test-cases are you referring to? Which one passes and which one fails?

kleopatra:
sry, forgot to repeat it here: should_click_on_button() fails because the button isn't clicked

kleopatra:
okay, a day's debugging later (fx11 only): 

-  a stand-alone version of awt.Robot passes the test in the core bug report, both for fx controls and swing components, so don't know why the AWTRobotAdapter used by default in BaseRobotImp doesn't work correctly (should it?)
- GlassRobotAdapter is not yet adjusted to fx11 (as you probably know better than I do :) so it might be expected to blow - as it does with a java.lang.NoClassDefFoundError: com/sun/glass/ui/Robot
-as PoC: a quick hack of glassRobotAdapter that's parameterized on GlassRobot (nothing for production, of course) works fine. Drawback: couldn't find a way to inject a custom impl, so c&p'ed the whole stack from ApplicationTest down to FxRobotContext just to replace BaseRobotImp to use the hacked adapter.

That's for today, hungry now <g>

brcolow:
[...]

PR #625, opened 2018-10-03
Make glass robot default on all platforms.

brcolow:
Investigating #581 

Locally on Windows 10 using the AWT robot the mouse goes to the wrong location (pixel and screen capture also fail, because those are the methods that can potentially use scaling).

But on Appveyor everything works as expected.

I am going to compare the screen device details as returned by AWT to see if there is a property we can use to try and differentiate when to scale or not, and use that instead of adding `PlatformAdapter.getOs() != OS.WINDOWS` condition to `AwtRobotAdapter.scaleRequired` (which is just for testing to see that this fixes things locally, but breaks Appveyor).

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,8 @@ you are using JUnit 4 in your project you would add a dependency on `testfx-juni
 </dependency>
 ```
 
-## Example: JUnit 4 with Hamcrest Matchers
+## Examples
+### JUnit 4 with Hamcrest Matchers
 
 ```java
 import org.testfx.framework.junit.ApplicationTest;
@@ -105,7 +106,7 @@ public class DesktopPaneTest extends ApplicationTest {
 }
 ```
 
-## Example: JUnit 4 with AssertJ Assertions
+### JUnit 4 with AssertJ Assertions
 
 ```java
 import org.testfx.framework.junit.ApplicationTest;
@@ -144,7 +145,7 @@ public class DesktopPaneTest extends ApplicationTest {
 }
 ```
 
-## Example: JUnit 5 with Hamcrest Matchers
+### JUnit 5 with Hamcrest Matchers
 
 ```java
 import org.testfx.framework.junit5.ApplicationTest;
@@ -177,7 +178,7 @@ class ClickableButtonTest {
 }
 ```
 
-## Example: JUnit 5 with AssertJ Assertions
+### JUnit 5 with AssertJ Assertions
 
 ```java
 import org.testfx.framework.junit5.ApplicationTest;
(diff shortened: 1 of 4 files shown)


brcolow:
Appveyor AWT display config:

```
  device: \Display0
    display mode: java.awt.DisplayMode@b20
    type: TYPE_RASTER_SCREEN
    is display change supported: false
    is full screen supported: true
    available accelerated memory: -1
    bounds: java.awt.Rectangle[x=0,y=0,width=1024,height=768]
    is multi buffer available: false
    color model: DirectColorModel: rmask=ff0000 gmask=ff00 bmask=ff amask=0
    color model pixel size: 24
    default transform: AffineTransform[[1.0, 0.0, 0.0], [0.0, 1.0, 0.0]]
    image capabilities (accelerated): false
    normalizing transform: AffineTransform[[1.333333333333333, 0.0, 0.0], [0.0, 1.333333333333333, 0.0]]
    is translucency capable: true
```

My local Windows 10 device config:

```
    device: \Display0
    display mode: 3440x1440x32bpp@50Hz
    type: TYPE_RASTER_SCREEN
    is display change supported: false
    is full screen supported: true
    available accelerated memory: -1048576
    bounds: java.awt.Rectangle[x=0,y=0,width=2752,height=1152]
    is multi buffer available: true
    color model: DirectColorModel: rmask=ff0000 gmask=ff00 bmask=ff amask=0
    color model pixel size: 24
    default transform: AffineTransform[[1.25, 0.0, 0.0], [0.0, 1.25, 0.0]]
    image capabilities (accelerated): true
    normalizing transform: AffineTransform[[1.666666666666667, 0.0, 0.0], [0.0, 1.666666666666667, 0.0]]
    is translucency capable: true
```

brcolow:
Another idea about how we could make this work automatically in all scenarios on Windows would be, when using the AWT robot, if a test fails, and scaling is enabled, see if it passes without scaling (or vice-versa). If that's the case automatically use scaling from then on with a warning message that informs the user that they are on a system that requires special casing scaling and recommend that they explicitly set `-Dtestfx.awt.scale=true` (a new property we could add that, when set, either scales or not and doesn't try the backup method). It's a little bit unwieldy, but so is trying to special case every CI provider that people may use TestFX with. At least this way it tries to intelligently do "the right thing".

Just thinking on this for now, won't make any hasty decisions.

brcolow:
The current approach, which I think is the simplest, is to make it so that AWT does *not* do scaling on Windows (for HiDPI) and thus when running on a local developer machine behavior will be as expected. If using the (headed) AWT robot on Appveyor with HiDPI (a pretty specific scenario although not *super* specific because currently headed glass robot deadlocks on Appveyor so AWT is the only  choice for headed testing) then the system property `testfx.awt.scale` must be set to `true`. Because of this new requirement I have added CI configuration (for Travis and Appveyor) to the README so that it is both more visible and (hopefully) kept more up-to-date than the wiki.

Issue #632, opened 2018-10-11
(testfx-junit5) ApplicationExtension.afterEach does not clean Mouse/Key events


arnobl:
## Expected Behavior

Using `ApplicationTest` provided by `testfx-junit`, the `internalAfter` method cleans the remaining events:
```java
    @AfterEach
    public final void internalAfter() throws Exception {
        // release all keys
        release(new KeyCode[0]);
        // release all mouse buttons
        release(new MouseButton[0]);
        FxToolkit.cleanupStages();
        FxToolkit.cleanupApplication(new ApplicationAdapter(this));
    }
```

## Actual Behavior

Using  `ApplicationExtension` provided by `testfx-junit5`, the `afterEach` method does not clean the remaining events:
```java
    @Override
    public void afterEach(ExtensionContext context) throws Exception {
        FxToolkit.cleanupApplication(new ApplicationAdapter(applicationFixture));
    }
```
To make it works as before I have to add in my test class:
```java
	@AfterEach
	void afterEach(final FxRobot robot) {
		robot.release(new KeyCode[0]);
		robot.release(new MouseButton[0]);
		WaitForAsyncUtils.waitForFxEvents(); // required to wait for the end of the UI events processing
	}
```
I guess these lines should be added at the beginning of `ApplicationExtension.afterEach`.

## Specifications

  - Version: 4.0.15-alpha
  - Platform: linux

brcolow:
@arnobl Would you have time to open a PR? If not I will do so, but you deserve the contribution 👍.

arnobl:
I will do that soon.

brcolow:
Fixed by https://github.com/TestFX/TestFX/pull/633.

PR #633, opened 2018-10-12
(fix) ApplicationExtension must free UI events after each test #632

arnobl:
Fix for #632 that affects the project testfx-junit5: `ApplicationExtension.afterEach` does not clean Mouse/Key events.

Tests are already provided by the test class `KeyAndButtonReleaseTest`, which is currently disabled:
https://github.com/TestFX/TestFX/blob/master/subprojects/testfx-junit5/src/test/java/org/testfx/framework/junit5/KeyAndButtonReleaseTest.java


brcolow:
Thanks for doing this. Do you have any thoughts about how we can force test ordering on Junit 5 so that we can make the tests not disabled? If not I will merge as-is.

arnobl:
This is a current opened issue in JUnit (https://github.com/junit-team/junit5/issues/48). No mainstream JUnit solution yet as as I as know.

brcolow:
Merged. Thanks for taking the initiative on this.

PR #642, opened 2018-11-20
Matcher for matching ListViews with given value in their selection.

htmoia:
This PR makes it possible to match ListViews by values in their selection model, i.e.
```java
assertThat("#my-list-view-id", ListViewMatchers.hasSelectedRow("selected value"));
````


htmoia:
Ok, fixed the intendation.

brcolow:
LGTM, ideally we would add the new matchers to ListViewAssert as well but I can do that as a follow-up.

Issue #507, opened 2018-01-17
NullPointerException on Empty cells


MikhailMS:
Hello there,

I'm testing my table view with your library, and currently discovered an issue with empty cells.
So, if I got a scene with two table views, and one of the tables got one row of data with only one cell to have value and others are empty (which is desired), and I'm calling `TableViewMatchers.hasTableCell` to see, if this row has a cell with a value, I'm getting NullPointerException from `TableViewMatchers.getRowValues`, as it seems like this method doesn't handle empty cells(which seems to be a Null as a default value for TableView).

brcolow:
Do you think you could make a simple example/test-case that demonstrates the current behavior versus expected behavior?

torakiki:
I'm also getting the exception
```
java.lang.NullPointerException
	at org.testfx.matcher.control.TableViewMatchers.getRowValues(TableViewMatchers.java:195)
	at org.testfx.matcher.control.TableViewMatchers.containsRowAtIndex(TableViewMatchers.java:173)
	at org.testfx.matcher.control.TableViewMatchers.lambda$containsRowAtIndex$6(TableViewMatchers.java:126)
	at org.testfx.matcher.base.GeneralMatchers$3.matchesSafely(GeneralMatchers.java:124)
```
In my case it's something [like this](https://stackoverflow.com/questions/33353014/creating-a-row-index-column-in-javafx) so a table column with a cellFactory but no cellValueFactory... `getRowValues` throws a NPE here `column.getCellValueFactory().call(cellDataFeatures)`

JanMosigItemis:
Hi, FYI I created a PR for this. Maybe it helps.

PR #650, opened 2019-02-06
507 npe on empty cells

JanMosigItemis:
Hi there, I had the problem of using "special" cells in a TableView. Those cells would display a button instead of a value. That seemed to cause NPEs in `TableViewMatchers.java:195`, because in this scenario no CellValueFactory is set.

I implemented a quick fix for that which may also "fix" issue #507. The TableViewMatchers's methods do now accept `null` as a cell "value", which is interpreted as "no value" or "no cell value factory set" respectively.

This comes with a minor change in semantics of `org.testfx.matcher.control.TableViewMatchers.containsRow(Object...)` and `org.testfx.matcher.control.TableViewMatchers.containsRowAtIndex(int, Object...)` which I have documented in the JavaDoc. Hope it's ok though.

What do you think?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,7 @@ screenshot*.png
 tags
 
 /subprojects/testfx-core/out/
+
+# Eclipse's gradle plugin
+bin/
+
--- a/subprojects/testfx-core/src/main/java/org/testfx/matcher/control/TableViewMatchers.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/matcher/control/TableViewMatchers.java
@@ -25,6 +25,7 @@ import java.util.Objects;
 import java.util.StringJoiner;
 import java.util.stream.Collectors;
 
+import javafx.beans.property.SimpleObjectProperty;
 import javafx.beans.value.ObservableValue;
 import javafx.scene.control.Cell;
 import javafx.scene.control.TableColumn;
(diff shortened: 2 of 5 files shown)


brcolow:
The idea makes sense to me. Thank you for opening a PR. There are quite a few linting errors (import order mostly) as we have pretty aggressive linting. All violations should be reported in the CI via Checkstyle. Once those are fixed this LGTM.

JanMosigItemis:
@brcolow I'll try to fix them. However, it is a bit hard to do work in Eclipse, because the gradle plugin complains about circular dependencies, i. e. the junit modules require core and core requires junit.

JanMosigItemis:
Seems like, the errors have been fixed. Travis build failed due to some non related timeout issues. I guess, those are flickering tests? Could you please have a look?

brcolow:
LGTM. Merged. Thanks very much - will look into the circular dependency issue.

brcolow:
Merged with clean commit message as https://github.com/TestFX/TestFX/commit/011dec6806a29b4abd51b0d05dddaa071dbeb93f

JanMosigItemis:
Your quick reaction is very much appreciated. 👍

brcolow:
Your contribution is also appreciated!

kleopatra:
hmm ... the fix is a bit fishy:

- the old code was wrong in assuming cellValueFactory != null, that's fixed by the new code
- the new code is (borderline) wrong in replacing a null cellObservableValue (which is a valid return value) with a fake observable (right now, can't think of a use-case where the distinction between "no observable" and "observable-with-null" might be important but suspect that it exists ;)

the error is in the upstream methods of TableViewMatchers: they don't handle a null observable,  f.i 

```
private static <T> boolean containsRowAtIndex(TableView<T> tableView, int rowIndex, Object...cells) {
    if (rowIndex < 0 || rowIndex >= tableView.getItems().size()) {
        return false;
    }

    List<ObservableValue<?>> rowValues = getRowValues(tableView, rowIndex);
    for (int i = 0; i < cells.length; i++) {
// ----------> this unguarded access is incorrect because table's cellObservableValue might be null
        if (rowValues.get(i).getValue() == null) {

```

A deeper fix would cleanup all the locations of unguarded access to a potential null.

JanMosigItemis:
@kleopatra sure it is and sure it may, but imo that's a different story. This PR is only about providing a workaround for the issue at hand. If you could come up with a PR this would be very appreciated.

PR #664, opened 2019-07-31
documentation fixes

frenchy64:


subprojects/testfx-core/src/main/java/org/testfx/api/FxToolkit.java
```diff
@@ -196,7 +196,6 @@ public static void setupFixture(Runnable runnable) throws TimeoutException {
 
     /**
      * Runs the given {@code callable} on the {@code JavaFX Application Thread} and returns once finished.
-     * before returning.
```
brcolow: 👍

subprojects/testfx-core/src/main/java/org/testfx/robot/Motion.java
```diff
@@ -38,7 +38,7 @@
  * <p>
  * Traveling in a straight-line between a and b (that is, tracing the hypotenuse) is
  * {@code DIRECT}. Traveling first from a to c and then from c to b is {@code HORIZONTAL_FIRST}.
- * Traveling first from a to d and then from d to b is {@code VERTICAL_FIRST}. {@code DIRECT}
+ * Traveling first from a to d and then from d to b is {@code VERTICAL_FIRST}. {@code DEFAULT}
```
brcolow: I am not sure about this change. In terms of the enum itself there is a `DEFAULT` which has the semantic meaning "Whatever TestFX defines as the default way to move between two point.". `DIRECT` is currently the `DEFAULT`, but that could change in the future. Having the value of `DIRECT` means specifically to use direct (naively diagonal, although of course it could be only horizontal or only vertical depending on the start and end points) motion whereas `DEFAULT` means use the TestFX default which is *currently* `DIRECT` but could change!

frenchy64: I'm confused. Are we on the same page that the docs currently say `DIRECT`, and this PR changes it to say `DEFAULT`?

brcolow: I was the one confused. Sorry about that, lol!

PR #669, opened 2019-08-07
[enhancement] Including module-info.java in the JAR artifacts.

brcolow:


brcolow:
Currently have some sort of split package, but not sure why:

`Module org.testfx contains package javafx.embed.swing, module javafx.swing exports package javafx.embed.swing to org.testfx`

PR #696, opened 2020-02-09
Sleep after atomic mouse actions

Thomas-Baeckeroot:
Without sleep after mouse actions, some OSes miss actions. This is especially true for double-click.
Human normally clicks in 50~ms, below 5ms it seems that click are missed. This change add a sleep of 20ms that is quick enough to be catch by OSes and closer from real human than nothing.
With this change, sleep after double-click in ClickRobotImpl becomes useless (less reliable than sleep after atomic mouse actions press/release).


brcolow:
Do you mind fixing the lint issues?

```
[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-core/src/main/java/org/testfx/robot/impl/MouseRobotImpl.java:38:11: 'static' modifier out of order with the JLS suggestions. [ModifierOrder]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-core/src/main/java/org/testfx/robot/impl/MouseRobotImpl.java:38:31: Name 'mouseActionDurationMillis' must match pattern '^[A-Z][A-Z0-9]*(_[A-Z0-9]+)*$'. [ConstantName]
```

Thanks for your PR!

Issue #700, opened 2020-03-30
proposal: new '@TestFx annotation and corresponding [Selective]JavaFxInterceptor extension to control thread execution.


RalphSteinhagen:
First, thanks a lot for this lib and making JavaFX more unit-testable!

This concerns whether `@Test`  functions are executed within our outside the JavaFX UI thread.

We wrote some small JUnit 5-compatible extensions that can -- in addition to the robot interface -- control whether the individual unit tests are executed within the JavaFX UI thread (ie. simple getter/setter consistency tests) or by one of the JUnit thread pool tasks. Usage example:
a) `@ExtendWith(JavaFxInterceptor.class)` before the test class:
 - functions annotated with `@TestFx` and/or `@Test` are executed within the JavaFX UI thread

b) `@ExtendWith(SelectiveJavaFxInterceptor.class)` before the test class:
 - only test functions annotated with `@TestFx` are executed within the JavaFX UI thread
 - functions with `@Test` are executed by JUnit's standard thread pool

Implementation and examples can be found at (Apache 2.0, no re-use restrictions):
https://github.com/GSI-CS-CO/chart-fx/tree/master/chartfx-chart/src/test/java/de/gsi/chart/ui/utils

We thought this might be also useful for other TestFX users. I hope that the new functionality makes the testing of JavaFX classes and the readability of the corresponding unit-tests more consistent for the different use cases.

Enjoy!

brcolow:
Would you be willing to submit a PR so we can add this to testfx-junit5?

RalphSteinhagen:
@brcolow am feeling a bit shy, also because I wouldn't have much else to contribute right now. 

For me some of the open design questions are:
* is having a utility class OK for TestFX?
* should this possibly be part of the 'ApplicationExtension'?
* is the `@TestFx` annotation OK with your design goals?
* is the configurable distinction between `@TestFx` and `@Test` using the two different extensions necessary or would just `@Test` always being executed in the JavaFX UI thread suffice.

The thing is that these design choices open your interface to many more possibilities. While I am personally happy to and prefer to provide such rich and diverse interface for our users within our projects, I also know that some other developers prefer to have a very limited scope of functionalities they'd like to provide to ease their long-term maintenance goals (ie. less code, less to maintain). Thus I feel it be better that the TestFX core developer take the decision whether this fits (as is) or needs to be modified to whatever target is preferred.

In any case, you are free to re-use whatever piece of the code and unit-tests you like and find useful.

[...]

PR #702, opened 2020-05-09
JUnit5 TestFX extension to execute @Test within JavaFX UI thread

RalphSteinhagen:
Simple JUnit 5 extension to ensure that @test statements are executed in the JavaFX UI thread. This is (strictly) necessary when testing setter and/or getter methods of JavaFX classes (ie. Node derived, properties, etc).

See also issue #700 and real-world usage in the following [example project](https://github.com/GSI-CS-CO/chart-fx/tree/master/chartfx-chart/src/test/java/de/gsi/chart)

Tests compile successfully with Maven but I am unfamiliar with Gradle and TestFX's use of it.
Presently org.junit.platform.commons.logging.Logger[Factory] module/visibility isn't properly exported. 

@brcolow -- since you kindly suggested this PR -- could you have a look/fix this Gradle issue?

Many thanks in advance. Much appreciated :+1: Hope this is useful also for others.

### DIFF
--- a/subprojects/testfx-junit5/testfx-junit5.gradle
+++ b/subprojects/testfx-junit5/testfx-junit5.gradle
@@ -58,6 +58,7 @@ afterEvaluate {
         testImplementation group: 'org.hamcrest', name: 'hamcrest', version: '2.1'
         implementation "org.assertj:assertj-core:3.13.2"
         //testImplementation "org.assertj:assertj-core:3.13.2"
+        testImplementation 'org.awaitility:awaitility:4.0.2'
 
         if (JavaVersion.current().isJava12Compatible()) {
             testCompile 'org.testfx:openjfx-monocle:jdk-12.0.1+2'
@@ -138,9 +139,11 @@ afterEvaluate {
                 options.compilerArgs = [
                         '--module-path', classpath.asPath,
                         '--add-modules', 'org.junit.jupiter.api',
+                        '--add-modules', 'org.junit.platform.commons',
                         '--add-modules', 'org.hamcrest',
                         '--add-modules', 'org.assertj.core',
                         '--add-reads', "$moduleName=org.junit.jupiter.api",
+                        '--add-reads', "$moduleName=org.junit.platform.commons",
                         '--add-reads', "$moduleName=javafx.controls",
                         '--add-reads', "$moduleName=org.hamcrest",
                         '--add-reads', "$moduleName=org.assertj.core",
--- a/subprojects/testfx-junit5/testfx-junit5.gradle
+++ b/subprojects/testfx-junit5/testfx-junit5.gradle
@@ -139,11 +139,9 @@ afterEvaluate {
                 options.compilerArgs = [
                         '--module-path', classpath.asPath,
                         '--add-modules', 'org.junit.jupiter.api',
-                        '--add-modules', 'org.junit.platform.commons',
                         '--add-modules', 'org.hamcrest',
                         '--add-modules', 'org.assertj.core',
                         '--add-reads', "$moduleName=org.junit.jupiter.api",
-                        '--add-reads', "$moduleName=org.junit.platform.commons",
                         '--add-reads', "$moduleName=javafx.controls",
                         '--add-reads', "$moduleName=org.hamcrest",
                         '--add-reads', "$moduleName=org.assertj.core",
--- a/gradle/check/checkstyle.xml
+++ b/gradle/check/checkstyle.xml
@@ -20,7 +20,10 @@
         "http://www.puppycrawl.com/dtds/configuration_1_2.dtd">
 <module name="Checker">
     <property name="charset" value="UTF-8"/>
-
+    <module name="LineLength">
+        <property name="max" value="120"/>
+        <property name="ignorePattern" value="^package.*|^import.*|a href|href|http://|https://|ftp://"/>
+    </module>
     <module name="TreeWalker">
         <module name="AvoidNestedBlocks">
             <property name="allowInSwitchCase" value="true"/>
@@ -42,10 +45,6 @@
             <property name="arrayInitIndent" value="2"/>
         </module>
         <module name="CommentsIndentation"/>
-        <module name="LineLength">
-            <property name="max" value="120"/>
-            <property name="ignorePattern" value="^package.*|^import.*|a href|href|http://|https://|ftp://"/>
-        </module>
         <module name="SeparatorWrap">
             <property name="tokens" value="DOT"/>
             <property name="option" value="nl"/>
--- a/subprojects/testfx-core/src/main/java/module-info.java
+++ b/subprojects/testfx-core/src/main/java/module-info.java
@@ -1,6 +1,6 @@
 /*
  * Copyright 2013-2014 SmartBear Software
- * Copyright 2014-2019 The TestFX Contributors
+ * Copyright 2014-2020 The TestFX Contributors
  *
  * Licensed under the EUPL, Version 1.1 or - as soon they will be approved by the
  * European Commission - subsequent versions of the EUPL (the "Licence"); You may
--- a/subprojects/testfx-core/src/main/java/org/testfx/api/FxAssert.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/api/FxAssert.java
@@ -1,6 +1,6 @@
 /*
  * Copyright 2013-2014 SmartBear Software
- * Copyright 2014-2019 The TestFX Contributors
+ * Copyright 2014-2020 The TestFX Contributors
  *
  * Licensed under the EUPL, Version 1.1 or - as soon they will be approved by the
  * European Commission - subsequent versions of the EUPL (the "Licence"); You may
--- a/subprojects/testfx-core/src/main/java/org/testfx/api/FxAssertContext.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/api/FxAssertContext.java
@@ -1,6 +1,6 @@
 /*
  * Copyright 2013-2014 SmartBear Software
- * Copyright 2014-2019 The TestFX Contributors
+ * Copyright 2014-2020 The TestFX Contributors
  *
  * Licensed under the EUPL, Version 1.1 or - as soon they will be approved by the
  * European Commission - subsequent versions of the EUPL (the "Licence"); You may
--- a/subprojects/testfx-core/src/main/java/org/testfx/api/FxRobot.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/api/FxRobot.java
@@ -1,6 +1,6 @@
 /*
  * Copyright 2013-2014 SmartBear Software
- * Copyright 2014-2019 The TestFX Contributors
+ * Copyright 2014-2020 The TestFX Contributors
  *
  * Licensed under the EUPL, Version 1.1 or - as soon they will be approved by the
  * European Commission - subsequent versions of the EUPL (the "Licence"); You may
--- a/subprojects/testfx-core/src/main/java/org/testfx/api/FxRobotContext.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/api/FxRobotContext.java
@@ -1,6 +1,6 @@
 /*
  * Copyright 2013-2014 SmartBear Software
- * Copyright 2014-2019 The TestFX Contributors
+ * Copyright 2014-2020 The TestFX Contributors
  *
  * Licensed under the EUPL, Version 1.1 or - as soon they will be approved by the
  * European Commission - subsequent versions of the EUPL (the "Licence"); You may
--- a/subprojects/testfx-core/src/main/java/org/testfx/api/FxRobotException.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/api/FxRobotException.java
@@ -1,6 +1,6 @@
 /*
  * Copyright 2013-2014 SmartBear Software
- * Copyright 2014-2019 The TestFX Contributors
+ * Copyright 2014-2020 The TestFX Contributors
  *
  * Licensed under the EUPL, Version 1.1 or - as soon they will be approved by the
  * European Commission - subsequent versions of the EUPL (the "Licence"); You may
--- a/subprojects/testfx-core/src/main/java/org/testfx/api/FxRobotInterface.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/api/FxRobotInterface.java
@@ -1,6 +1,6 @@
 /*
  * Copyright 2013-2014 SmartBear Software
- * Copyright 2014-2019 The TestFX Contributors
+ * Copyright 2014-2020 The TestFX Contributors
  *
  * Licensed under the EUPL, Version 1.1 or - as soon they will be approved by the
  * European Commission - subsequent versions of the EUPL (the "Licence"); You may
--- a/subprojects/testfx-core/src/main/java/org/testfx/api/FxService.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/api/FxService.java
@@ -1,6 +1,6 @@
 /*
  * Copyright 2013-2014 SmartBear Software
- * Copyright 2014-2019 The TestFX Contributors
+ * Copyright 2014-2020 The TestFX Contributors
  *
  * Licensed under the EUPL, Version 1.1 or - as soon they will be approved by the
  * European Commission - subsequent versions of the EUPL (the "Licence"); You may
--- a/subprojects/testfx-core/src/main/java/org/testfx/api/FxServiceContext.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/api/FxServiceContext.java
@@ -1,6 +1,6 @@
 /*
  * Copyright 2013-2014 SmartBear Software
- * Copyright 2014-2019 The TestFX Contributors
+ * Copyright 2014-2020 The TestFX Contributors
  *
  * Licensed under the EUPL, Version 1.1 or - as soon they will be approved by the
  * European Commission - subsequent versions of the EUPL (the "Licence"); You may
--- a/subprojects/testfx-core/src/main/java/org/testfx/api/FxToolkit.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/api/FxToolkit.java
@@ -1,6 +1,6 @@
 /*
  * Copyright 2013-2014 SmartBear Software
- * Copyright 2014-2019 The TestFX Contributors
+ * Copyright 2014-2020 The TestFX Contributors
  *
  * Licensed under the EUPL, Version 1.1 or - as soon they will be approved by the
  * European Commission - subsequent versions of the EUPL (the "Licence"); You may
--- a/subprojects/testfx-core/src/main/java/org/testfx/api/FxToolkitContext.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/api/FxToolkitContext.java
@@ -1,6 +1,6 @@
 /*
  * Copyright 2013-2014 SmartBear Software
- * Copyright 2014-2019 The TestFX Contributors
+ * Copyright 2014-2020 The TestFX Contributors
  *
  * Licensed under the EUPL, Version 1.1 or - as soon they will be approved by the
  * European Commission - subsequent versions of the EUPL (the "Licence"); You may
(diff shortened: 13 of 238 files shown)


brcolow:
I will try and get some time to look at this. I really appreciate the PR. Getting Jigsaw modules to work without dedicated support in Gradle (when I wrote the Jigsaw stuff) has been a nightmare. I think JPMS support as improved as of late, it requires re-doing it from scratch as it's an absolute mess.

brcolow:
Really need to use something like https://github.com/java9-modularity/gradle-modules-plugin at a minimum or switch to Maven which has better JPMS support.

RalphSteinhagen:
@brcolow thanks for your help, especially since you worked with TestFX's build system before.

We normally use Maven, but I managed to get the Jigsaw module-ing to work for a simple Gradle setup (see [link](https://github.com/GSI-CS-CO/chart-fx/issues/161#issuecomment-617687951)), However, I was not able to transfer these experiences to the particular TestFX's Gradle config.

My hunch is that this is probably a one-liner but I cannot find an easy solution w/o rewriting the config which in turn probably would trigger lots of other side-effects. Thanks again for your help!

RalphSteinhagen:
@brcolow any chance of getting help for this? 

N.B. We are trying to move our [little lib](https://github.com/GSI-CS-CO/chart-fx/pull/160) also to Jigsaw and this/getting TestFX Jigsaw-compatible is an important puzzle piece.

brcolow:
So currently there are many linting failures. That is the first step to getting this merged:

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/JavaFXInterceptorUtils.java:32: Line is longer than 120 characters (found 127). [LineLength]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/JavaFXInterceptorUtils.java:41: Line matches the illegal pattern '^\s*\*\s*@author'. [RegexpSingleline]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/JavaFXInterceptorUtils.java:45:1: File contains tab characters (this is the first instance). [FileTabCharacter]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/JavaFXInterceptorUtils.java:46: Line is longer than 120 characters (found 135). [LineLength]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/JavaFXInterceptorUtils.java:49: Javadoc comment at column 10 has parse error. Missed HTML close tag 'code'. Sometimes it means that close tag missed for one of previous tags. [AtclauseOrder]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/JavaFXInterceptorUtils.java:59: Line is longer than 120 characters (found 122). [LineLength]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/JavaFXInterceptorUtils.java:68: Line is longer than 120 characters (found 138). [LineLength]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/JavaFXInterceptorUtils.java:80: Line matches the illegal pattern '^\s*\*\s*@author'. [RegexpSingleline]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/JavaFXInterceptorUtils.java:95:17: '}' at column 17 should be alone on a line. [RightCurly]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/JavaFXInterceptorUtils.java:108: Line is longer than 120 characters (found 135). [LineLength]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/JavaFXInterceptorUtils.java:111: Javadoc comment at column 10 has parse error. Missed HTML close tag 'code'. Sometimes it means that close tag missed for one of previous tags. [AtclauseOrder]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/JavaFXInterceptorUtils.java:130: Line is longer than 120 characters (found 139). [LineLength]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/JavaFXInterceptorUtils.java:142: Line matches the illegal pattern '^\s*\*\s*@author'. [RegexpSingleline]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/JavaFXInterceptorUtils.java:165:17: '}' at column 17 should be alone on a line. [RightCurly]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/TestFx.java:29: Line is longer than 120 characters (found 131). [LineLength]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/TestFx.java:31: Line matches the illegal pattern '^\s*\*\s*@author'. [RegexpSingleline]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/utils/FXUtils.java:32: Import statement for 'javafx.application.Platform' is in the wrong order. Should be in the 'STANDARD_JAVA_PACKAGE' group, expecting not assigned imports on this line. [CustomImportOrder]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/utils/FXUtils.java:33: Import statement for 'javafx.scene.Scene' is in the wrong order. Should be in the 'STANDARD_JAVA_PACKAGE' group, expecting not assigned imports on this line. [CustomImportOrder]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/utils/FXUtils.java:39: Line matches the illegal pattern '^\s*\*\s*@author'. [RegexpSingleline]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/utils/FXUtils.java:61: Line matches the illegal pattern '^\s*\*\s*@author'. [RegexpSingleline]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/utils/FXUtils.java:62: Line matches the illegal pattern '^\s*\*\s*@author'. [RegexpSingleline]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/utils/FXUtils.java:76: Line matches the illegal pattern '^\s*\*\s*@author'. [RegexpSingleline]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/utils/FXUtils.java:77: Line matches the illegal pattern '^\s*\*\s*@author'. [RegexpSingleline]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/utils/FXUtils.java:90: Line matches the illegal pattern '^\s*\*\s*@author'. [RegexpSingleline]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/utils/FXUtils.java:91: Line matches the illegal pattern '^\s*\*\s*@author'. [RegexpSingleline]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/utils/FXUtils.java:135:13: '}' at column 13 should be alone on a line. [RightCurly]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/utils/FXUtils.java:174:17: '}' at column 17 should be alone on a line. [RightCurly]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/utils/FXUtils.java:185:9: '}' at column 9 should be alone on a line. [RightCurly]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/utils/FXUtils.java:211:9: '}' at column 9 should be alone on a line. [RightCurly]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/utils/FXUtils.java:213:9: '}' at column 9 should be alone on a line. [RightCurly]

[ant:checkstyle] [ERROR] /home/travis/build/TestFX/TestFX/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/utils/FXUtils.java:219:9: '}' at column 9 should be alone on a line. [RightCurly]

Once those are fixed we can see the errors in CI.

brcolow:
With respect to the `org.junit.platform.commons.logging` package/module I see the following:

```
Internal logging package.
DISCLAIMER
These classes are intended solely for usage within the JUnit framework itself. Any usage by external parties is not supported. Use at your own risk!
```

This is an internal logging facade that Junit 5 uses. So far in the project we have not introduced logging and simply use `System.out.println`. If we want to introduce a logging library we should not depend on Junit 5's repackaged JUL facade but an actual logging library.

Because  `org.junit.platform.commons.logging`  is essentially a "private API" one would need the add the following compile-time argument:

`--add-exports org.junit.platform.commons/org.junit.platform.commons.logging=org.testfx.junit5` which is something ugly we definitely want to avoid if we can.

RalphSteinhagen:
@brcolow I presume we agree -- without delving into details -- `System.[out, err].println(...)` should not be used in production code and even less in a library unless the aim is to deliberately annoy end-users and developers with obfuscating printouts that cannot be easily filtered or suppressed. :smirk:

If JUnit5's logging facade should not be used, may I suggest to add [SLF4J](https://github.com/qos-ch/slf4j) as a dependency to TestFX? Version `2.0.0-alpha0` can be considered as stable. 
It's also a simple and widely used logging facade that can be rerouted to the console and/or nearly all other logging infrastructure services.

In case the move to SLF4J rocks the boat (and referring to the original issue) how shall we proceed? Shall I remove all logging (incl. 'warning') statements?

brcolow:
For the purpose and scope of this PR let's either comment out the logging statements or switch to System.out.println. We can revisit logging in a later, specific, PR. Don't want this good functionality to get side-tracked on an essentially side issue.

RalphSteinhagen:
> Execution failed for task ':testfx-core:licenseMain' :roll_eyes:

RalphSteinhagen:
> com.puppycrawl.tools.checkstyle.api.CheckstyleException: cannot initialize module TreeWalker - TreeWalker is not allowed as a parent of LineLength Please review 'Parent Module' section for this Check in web documentation if Check is standard.
	at com.puppycrawl.tools.checkstyle.Checker.setupChild(Checker.java:482)
	at com.puppycrawl.tools.checkstyle.api.AutomaticBean.configure(AutomaticBean.java:201)
	at org.infernus.idea.checkstyle.service.cmd.OpCreateChecker.execute(OpCreateChecker.java:61)
	at org.infernus.idea.checkstyle.service.cmd.OpCreateChecker.execute(OpCreateChecker.java:26)
	at org.infernus.idea.checkstyle.service.CheckstyleActionsImpl.executeCommand(CheckstyleActionsImpl.java:130)
	at org.infernus.idea.checkstyle.service.CheckstyleActionsImpl.createChecker(CheckstyleActionsImpl.java:60)
	at org.infernus.idea.checkstyle.service.CheckstyleActionsImpl.createChecker(CheckstyleActionsImpl.java:51)
	at org.infernus.idea.checkstyle.checker.CheckerFactoryWorker.run(CheckerFactoryWorker.java:46)
Caused by: com.puppycrawl.tools.checkstyle.api.CheckstyleException: TreeWalker is not allowed as a parent of LineLength Please review 'Parent Module' section for this Check in web documentation if Check is standard.
	at com.puppycrawl.tools.checkstyle.TreeWalker.setupChild(TreeWalker.java:137)
	at com.puppycrawl.tools.checkstyle.api.AutomaticBean.configure(AutomaticBean.java:201)
	at com.puppycrawl.tools.checkstyle.Checker.setupChild(Checker.java:477)
	... 7 more

RalphSteinhagen:
... there are still 9 unit-tests failing in other (untouched) TestFX submodules. I think I will leave these as is.... someone else needs to fix the broken CI/CD.

brcolow:
Is it the case that this PR introduces JavaFX API that is only available in 11 and newer? Specifically:

 Platform.requestNextPulse
 Scene#addPostLayoutPulseListener

RalphSteinhagen:
> Is it the case that this PR introduces JavaFX API that is only available in ~11~ 9 and newer? Specifically:
> 
> Platform.requestNextPulse
> Scene#addPostLayoutPulseListener

The old alternative is `Toolkit.getToolkit().requestNextPulse();` which works for Java 1.8 but is not portable/modularizable for JavaFX > Java 9.0 (private JDK API). 

You can find a JDK8 compatible port [here](https://github.com/GSI-CS-CO/chart-fx/blob/JDK8/chartfx-chart/src/main/java/de/gsi/chart/utils/FXUtils.java#L141-L212).

RalphSteinhagen:
> So far we have kept TestFX Java 8 compatible. Maybe it is time to progress...

We are also still supporting JDK8 strictly as legacy feature but it"s getting increasingly harder -- also because there are some bugs and less efficient implementations/APIs in the JDK8's JavaFX version that won't get fixed/back-ported. I think if people use the new JUnit5 API, JDK11 is also a safe assumption.

I read somewhere that the non-commercial support is supposed to be phased-out by end of 2020.

subprojects/testfx-junit5/testfx-junit5.gradle
```diff
@@ -138,9 +139,11 @@ afterEvaluate {
                 options.compilerArgs = [
                         '--module-path', classpath.asPath,
                         '--add-modules', 'org.junit.jupiter.api',
+                        '--add-modules', 'org.junit.platform.commons',
```
brcolow, line 142: Let's remove these now that we have made this PR not change logging stuff.

subprojects/testfx-junit5/testfx-junit5.gradle
```diff
@@ -138,9 +139,11 @@ afterEvaluate {
...
                         '--add-modules', 'org.hamcrest',
                         '--add-modules', 'org.assertj.core',
                         '--add-reads', "$moduleName=org.junit.jupiter.api",
+                        '--add-reads', "$moduleName=org.junit.platform.commons",
```
brcolow, line 146: Same here.

brcolow:
Once the two comments are addressed I will merge this. I don't want to duplicate/take-away the work you did with fixing license headers and checkstyle in my own PR. I will merge this and then work *from there* to fix CI.

subprojects/testfx-junit5/testfx-junit5.gradle
RalphSteinhagen: addressed by 75981dd

subprojects/testfx-junit5/testfx-junit5.gradle
RalphSteinhagen: addressed by 75981dd

brcolow:
Merged. Thanks a ton for your patience and hard work. I will work off this PR to fix the CI as you already addressed a few outstanding issues.

brcolow:
@RalphSteinhagen Just a note I am seeing the following failures on Mac TravisCI:

FXUtilsTests > testWithinFxThread() FAILED

    org.awaitility.core.ConditionTimeoutException: Condition with lambda expression in org.testfx.framework.junit5.utils.FXUtilsTests was not fulfilled within 200 milliseconds.

        at org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:165)

        at org.awaitility.core.CallableCondition.await(CallableCondition.java:78)

        at org.awaitility.core.CallableCondition.await(CallableCondition.java:26)

        at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:895)

        at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:864)

        at org.testfx.framework.junit5.utils.FXUtilsTests.testWithinFxThread(FXUtilsTests.java:90)

FXUtilsTests > testOutsideFxThread() STARTED

FXUtilsTests > testOutsideFxThread() FAILED

    org.awaitility.core.ConditionTimeoutException: Condition with lambda expression in org.testfx.framework.junit5.utils.FXUtilsTests was not fulfilled within 300 milliseconds.

        at org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:165)

        at org.awaitility.core.CallableCondition.await(CallableCondition.java:78)

        at org.awaitility.core.CallableCondition.await(CallableCondition.java:26)

        at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:895)

        at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:864)

        at org.testfx.framework.junit5.utils.FXUtilsTests.testOutsideFxThread(FXUtilsTests.java:141)

RalphSteinhagen:
@brcolow didn't it work before also on iOS? Unfortunately, I do not have a Mac and thus cannot reproduce this locally. However, I had a look at the stack traces you provided and tried to figure out as much as I can. Hope the following is helpful in isolating the problem.

I presume that you verified this error to be systematic and not due to TravisCI's iOS worker being overloaded. Regarding the latter, we found that this often causes trouble with timing-sensitive tests but restarting usually cured these problems due to temporarily stalling workers. 

What is good and a bit weird at the same time is that only one of the later tests failed in those functions.

> [..] ```at org.testfx.framework.junit5.utils.FXUtilsTests.testWithinFxThread(FXUtilsTests.java:90)```

The core functionality seems to work and the test method is being executed within the JavaFX thread as requested -- which is good :+1: -- but appears to fail and be blocked at `FXUtils.waitForFxTicks(testLabel.getScene(), 3);` :-1: which boils down to the following [JavaFX function call](https://github.com/brcolow/TestFX/blob/23c35d58492b0a349402403f068c54deba68c048/subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/utils/FXUtils.java#L151-L156) -- provided it is not an unlikely Awaitility problem for Mac:
```Java
        if (Platform.isFxApplicationThread()) {
            for (int i = 0; i < nTicks; i++) {
                Platform.requestNextPulse();
            }
            return true;
        }
```
No locks nor thread magic involved...a plain simple call to `Platform.requestNextPulse();` which in turn sets an AtomicBoolean flag of the quantum renderer to indicate that a tick should be computed which is not executed.

Similarly for:
> [..] ```at org.testfx.framework.junit5.utils.FXUtilsTests.testOutsideFxThread(FXUtilsTests.java:141)```

which also boils down to a call to `FXUtils.waitForFxTicks(testLabel.getScene(), 3);` and eventually an apparently blocking `Platform.requestNextPulse();` if I am not mistaken (again... no Mac). In this case, the test method is definitely run outside the JavaFX thread as this is being tested in the same method before.

I saw some activity regarding Monocle (which TestFX relies upon) and iOS on the openjdk/jfx issue mailing lists. Could this be related? Could you try bumping the monocle/JFX versions -- just for testing this issue?

Hope above explanations help a bit. Let me know if you found the culprit. Worst case, one could disable these few unit tests for iOS.

VirtualTim:
@brcolow would it be possible to have a new build published to maven that includes this change?

PR #706, opened 2020-06-23
Fix header level of Gradle and Maven subsections in README

Marcono1234:
The "Test Framework" and "Matcher/Assertions Library" sections are intended to be subsections of "Gradle" and "Maven". However, previously they were at the same header level as "Gradle" and "Maven".

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,10 +59,10 @@ javafx {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


brcolow:
Awesome. Thanks very much.

PR #707, opened 2020-06-23
Use HTTPS for links and update them in README

Marcono1234:
Changes:
- Use HTTPS for links where possible. spockframework.org and hamcrest.org are apparently incorrectly configured and cause a warning when accessed using HTTPS
- Update AssertJ link
- Use Maven Central group instead of search link


brcolow:
Awesome. Thanks very much.

PR #710, opened 2020-08-10
(docs) Note that we now only support Java 9 and greater.

brcolow:


brcolow:
Note to self:

```gradle
 modularity.inferModulePath = true
```

mvsoder:
Current TestFX will maintain compatibility with JDK 8

Issue #709, opened 2020-07-17
OutOfMemoryException by multiple tests - Scene is not cleared in the JavafxRobotAdapter


j-ferreira:
When using the write function in the FxRobot class, you will run into an OutOfMemoryExcpetion if you run multiple tests. 

After some heap dump analysis I found out that the scene of the JavaFX application is set in the JavafxRobotAdapter when the write method is called. But the scene remains in the JavafxRobotAdapter even after the test.

For multiple tests, this means that the scene is stored in memory several times. This allows you to run into an OutOfMemoryException. 

In our current project our JavaFX application is about 300MB in memory and we have about 60 tests. Therefore we ran into an OutOfMemoryException sometime. 
Currently we have found a workaround. We create a small stage at the end of each test and on this stage we call the write method to overwrite our actual stage in the JavafxRobotAdapter to free memory.

Additionally I noticed during the analysis in the heap dump that the complete FxRobotContext for each test is kept during the whole runtime. Is this really necessary? 
Although we have found this workaround now, the amount of memory required increases from test to test, because the FxRobotContext is kept for each test even though the test has already run.

If you need an example, I can upload a small example.

## Specifications

  - Version: testfx-4.0.16-alpha
  - Platform: JavaFX 13, AdoptOpenJDK 11, Windows 10

brcolow:
If you have the time and/or motivation to submit a PR it would be super appreciated. We have heard of this issue for quite some time now and it is a real pain point.

j-ferreira:
I'm not really familiar with the architecture of TestFX now, but I can try it. I just found this out over the heap dumps and without background knowledge. 
But what would the target be? That the Scene in the JavafxRobotAdapter is set to null at the end of each test or that the FxRobotContext of the corresponding test is deleted after the test is finished?

I think deleting the FxRobotContext would free up much more memory, but I can't really estimate the effort and I don't know if the FxRobotContext is still needed after the corresponding test?

brcolow:
These are great questions and I think can only really been answered with trial and error. The original author of TestFX is @hastebrot  - he is possibly the only one that could guide you on this work. My contributions have been more geared towards maintenance (and I've done a bad job at that lately).

Issue #689, opened 2019-09-18
Remove Application from ParametersImpl


qtran:
After we ported our tests from 3.x to 4.0.15 we are running sometimes into memory problems.
I think the problem is that the Application is registered each time in ParametersImpl and is never removed. 
So the GC can never remove the test classes from the memory.
Our current workaround is to remove the Application via Reflection from ParametersImpl:

```
public static void clearApplication() {
	try {
		final Field f = ParametersImpl.class.getDeclaredField("params");
		f.setAccessible(true);

		final Map<Application, Parameters> params = (Map<Application, Parameters>) f.get(null); 
		params.clear();
	} catch (NoSuchFieldException | SecurityException | IllegalArgumentException | IllegalAccessException e) {
	}
}
```

Running with JUnit 4, Java 8

brcolow:
Would you be open to submitting a pull request?

PR #724, opened 2021-05-02
(fix) Unregister scene in type keyboard.

AndFr:
Scene will be unregistered after an keyboard input was made.
This will probably fix the error described in https://github.com/TestFX/TestFX/issues/709

### DIFF
--- a/subprojects/testfx-core/src/main/java/org/testfx/robot/impl/BaseRobotImpl.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/robot/impl/BaseRobotImpl.java
@@ -75,6 +75,7 @@ public class BaseRobotImpl implements BaseRobot {
         javafxRobotAdapter.keyPress(key);
         javafxRobotAdapter.keyType(KeyCode.UNDEFINED, character);
         javafxRobotAdapter.keyRelease(key);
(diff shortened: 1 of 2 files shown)


hastebrot:
@AndFr Moin! Thank you for this pull request!

Do you think this can cause any performance problems? Should we put this behind a configuration option?

AndFr:
@hastebrot Moin! I've provided a minimal testfx project to demonstrate the memory leak behavior.
https://github.com/AndFr/TestFXLeakFinder

I am sure that this is not the cure for all problems, but in this particular case it could solve a problem.

AndFr:
I don't have any experience in working on open source projects, so were the failed checks cause by my changed or is there some kind of problem in the build pipeline ?

mvsoder:
@AndFr, thank you for your minimal project to show this situation. I have reproduced your results and will be researching the causes and solutions.

mvsoder:
@AndFr, again, thank you for your project code to show this situation. I think I have it sorted out and this PR is no longer necessary. After also reading issue https://github.com/TestFX/TestFX/issues/689 I was able to put a solution together. Feel free to review https://github.com/TestFX/TestFX/pull/774.

Issue #722, opened 2021-04-10
Any chance for JPEG snapshot images?


io7m:
## Expected Behavior
Currently, the image capture functionality unconditionally writes images in PNG format:

https://github.com/TestFX/TestFX/blob/c80fa9ca5e7852f8a8ecf71286b1cd5cc07836dc/subprojects/testfx-core/src/main/java/org/testfx/service/support/impl/CaptureSupportImpl.java#L124

I upload quite a lot of screenshots via GitHub Actions so that I can diagnose test failures. While the individual images aren't very big as PNG files, the total size quickly adds up as the number of images grows. Given that these are strictly diagnostic images, I could save quite a bit of time downloading if the images were low-quality lossy JPEG images.

## Actual Behavior

Add an option to allow me to specify the captured image format.

## Specifications

  - Version: 4.0.16-alpha
  - Platform: Linux sunflower 5.11.7-arch1-1 #1 SMP PREEMPT Wed, 17 Mar 2021 16:59:58 +0000 x86_64 GNU/Linux

brcolow:
PR welcome. Being able to choose the image format would be neat.

PR #726, opened 2021-05-13
(feature) Add the option to export an image to different filetype than png.

AndFr:
Added the option to save images to other types than PNG.

Its now possible to save images to
JPG, PNG, BMP, GIF, TIF

This issue was mentioned here:
https://github.com/TestFX/TestFX/issues/722


mvsoder:
This will need to be rebased (updated to current master) before it will build.

PR #729, opened 2021-05-16
Old year in license header cause problem

AndFr:
Seems that the wrong year caused problems in the build pipeline. I've updated the year in all files.


brcolow:
Thank you :)

PR #741, opened 2022-04-25
Fixed license header year

apollox:
The wrong year caused problems in the build pipeline


koppor:
It is 2023 now 😅

svenruppert:
IS fixed with current year 2023 - Cheers

PR #758, opened 2023-06-01
[BREAKING] (chore) Enable GitHub Actions and Upgrade to Gradle 8.1.1.

mvsoder:
This rather large pull requests enables GitHub Actions for an initial build environment (Ubuntu 20.04 using JDK 11) as well as upgrades to Gradle 8.1.1. This PR also includes the changes from #757. More updates to come, this is just the beginning. For an idea what this look like, please see the builds in action here: https://github.com/marksoderquist/testfx/actions.


hastebrot:
Hi, thank you very much for this pull request.

I guess this will automatically activate the github actions for pull request branches and the main branch with JDK 11 builds succeeding and JDK 17 failing?

mvsoder:
This PR will automatically enable GitHub actions, but only enables the build for JDK 11 because the build for JDK 17 is still not passing. I prefer that the main branch always stay in a passing state. Once I figure out how to deal with JDK 17, I'll submit another PR.

mvsoder:
I also recommended merging this PR before merging others so we can take advantage of the build and tests for merging the other PRs.

subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/ApplicationExtension.java
```diff
@@ -119,7 +119,7 @@ private Method validateStopMethod(Method stopMethod) {
     }
 
     private void setField(Object instance, Field field, Object val) throws IllegalAccessException {
-        boolean wasAccessible = field.isAccessible();
+        boolean wasAccessible = field.canAccess(instance);
```
hastebrot, line 122: note to myself:

[BREAKING] `AccessibleObject.canAccess()` requires Java 9.

hastebrot:
I declare this as breaking change since the change to `subprojects/testfx-junit5/src/main/java/org/testfx/framework/junit5/ApplicationExtension.java` requires Java 9 and `README.md` states `Support for: Java 8/9/10/11+`. However I think we can declare `junit5` plugin requires Java 9, and remove the `BREAKING`.

hastebrot:
So I'm not sure. Travis CI and AppVoyer seem to be broken and I can't remember if we ran CI always on Java 8. I guess when we update to Java 11 and also run CI on it, we have to drop Java 8 support.

mvsoder:
@hastebrot "`Field.isAccessible()` still exists in Java SE 18 and 20, even when it is deprecated." This is true, however, according to the JavaDoc( https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/reflect/AccessibleObject.html#isAccessible() ), it incorrectly infers access to the `accessible` flag when it actually refers to something slightly different. From the JavaDoc:

"This method is deprecated because its name hints that it checks if the reflected object is accessible when it actually indicates if the checks for Java language access control are suppressed. This method may return false on a reflected object that is accessible to the caller. To test if this reflected object is accessible, it should use [canAccess(Object)](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/reflect/AccessibleObject.html#canAccess(java.lang.Object))."

Thus the change from `isAccessible()` to `canAccess()`.

mvsoder:
Furthermore, this change does make the code base incompatible with Java 8. In talking with @svenruppert, we agree that we need to support the LTS releases 11 and 17, but we wonder about Java 8. Keeping the code compatible with Java 8 complicates the build process and limits the advances of Java 11 and Java 17. Along those lines, Java 21 should be the next LTS and it should be arriving in just a few months. Thoughts on Java 8 support?

koppor:
I am happy to see that there is activity in TestFX! I am very much in favor to base on JDK 17 or later. This is in line with JavaFX, where TestFX is build for. I checked https://openjfx.io/openjfx-docs/. Latest JavaFX requires JDK 17:

> In any case, for both options, it is required to have a recent version of JDK 20, or at least JDK 17. 

Thus, it is IMHO perfectly OK if TestFX also follows this requirement.

In addition, this should make contributing to TestFX easier.

subprojects/testfx-spock/testfx-spock.gradle
```diff
@@ -16,83 +16,80 @@
...
-    final String osName = System.getProperty("os.name").toLowerCase()
-    if (osName.contains("linux")) {
-        return ("linux")
-    } else if (osName.contains("mac os x") || osName.contains("darwin") || osName.contains("osx")) {
-        return ("mac")
-    } else if (osName.contains("windows")) {
-        return ("win")
-    }
-    return ""
+	final String osName = System.getProperty("os.name").toLowerCase()
```
koppor, line 21: Is it really necessary to replace spaces by tabs? I did not see any `.editorconfig` here and therefore I wonder why this was replaced.

mvsoder: Nope. Not necessary. This was an oversight.

Issue #764, opened 2023-06-20
(chore) JavaDoc generation fails with invalid flag html5


mvsoder:
## Expected Behavior
JavaDoc tool generates the JavaDocs during the build

## Actual Behavior
JavaDoc tool fails with: `javadoc: error - invalid flag: -html5`

To get the fastest possible support, create a [minimal example](https://stackoverflow.com/help/mcve)
and upload it to GitHub.

## Specifications

  - Version: TestFX 4.0.17-alpha
  - Platform: Linux Mint 21 (Ubuntu 20.04)

mvsoder:
This is because the JavaDoc 8 tools did not support html5.

PR #765, opened 2023-06-20
(fix for TestFX 4 JDK 8 support) Fix JavaDoc invalid flag html5.

mvsoder:
Fix for #764


mvsoder:
Required for TestFX 4 JDK 8 compatibility.

PR #772, opened 2023-06-21
README update name, badges and requirements

mvsoder:


brcolow:
Great to see TestFX alive again. Hope not all of my changes make for a higher maintenance burden. Thanks for stepping up!

PR #774, opened 2023-06-22
Fix memory leak due to ParametersImpl

### LINKED ISSUE #689: Remove Application from ParametersImpl

mvsoder:
Due to the implementation of ParametersImpl, there was a memory leak if tests started the application over and over. This PR implements a solution to clean up the application parameters stored in ParametersImpl and implements a test to help ensure we maintain good memory health.

Fixes #689 and #724

### DIFF
--- a/subprojects/testfx-core/src/main/java/org/testfx/toolkit/impl/ToolkitServiceImpl.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/toolkit/impl/ToolkitServiceImpl.java
@@ -178,7 +178,8 @@ public class ToolkitServiceImpl implements ToolkitService {
             Map<Application, Application.Parameters> params;
             params = (Map<Application, Application.Parameters>) field.get(null);
             params.remove(application);
-        } catch (Exception exception) {
+        }
+        catch (Exception exception) {
             exception.printStackTrace();
         }
     }
--- a/subprojects/testfx-core/src/main/java/org/testfx/toolkit/impl/ToolkitServiceImpl.java
+++ b/subprojects/testfx-core/src/main/java/org/testfx/toolkit/impl/ToolkitServiceImpl.java
@@ -17,9 +17,11 @@
 package org.testfx.toolkit.impl;
 
 import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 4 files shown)


AndFr (approved):
Seems to be ok from my POV

subprojects/testfx-core/src/main/java/org/testfx/toolkit/impl/ToolkitServiceImpl.java
svenruppert, line 1: Here is no Assert.

subprojects/testfx-core/src/test/java/org/testfx/cases/issue/MemoryLeakBug.java
svenruppert, line 1: Here is no assert - this test would never fail..

mvsoder: The test would fail due to an OutOfMemoryError being thrown if there were a memory leak.

mvsoder: I could add an assert that we have not had "large" memory growth by the end of the test if that is desired.

mvsoder: Added both a memory growth check and notes about the failure mode

subprojects/testfx-core/src/main/java/org/testfx/util/WaitForAsyncUtils.java
svenruppert, line 1: extract classnames as constants, please - "com.sun.javafx.tk.quantum.PaintCollector"

mvsoder: Sure thing

svenruppert (approved):
LGTM

PR #777, opened 2023-06-23
Remove Gradle references to jcenter and bintray

mvsoder:
One of several cleanup tasks for Gradle


DavidGliddonNZ (approved):
Mkae sense with jcenter service being shut down

svenruppert (approved):

PR #782, opened 2023-07-07
Improve exception message when missing libgtk-3-0

mvsoder:


subprojects/testfx-core/src/main/java/org/testfx/api/FxToolkit.java
```diff
@@ -260,4 +266,20 @@ public static boolean isFXApplicationThreadRunning() {
         return false;
     }
 
+    static void handleCommonRuntimeExceptions(RuntimeException exception) {
+        if (exception.getCause() instanceof UnsupportedOperationException) {
+            UnsupportedOperationException unsupportedOperationException =
+                    (UnsupportedOperationException)exception.getCause();
+            if (unsupportedOperationException.getMessage().equalsIgnoreCase("Internal Error")) {
+                String className = unsupportedOperationException.getStackTrace()[0].getClassName();
+                if (className.startsWith("com.sun.glass.ui.gtk.GtkApplication")) {
```
svenruppert, line 275: Extract in const , please.

subprojects/testfx-core/src/test/java/org/testfx/api/FxToolkitTest.java
```diff
@@ -0,0 +1,39 @@
...
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+public class FxToolkitTest {
+
+    @Test
+    public void improve_common_runtime_exception_messages() {
+        // given
+        StackTraceElement element = new StackTraceElement("com.sun.glass.ui.gtk.GtkApplication", "", "", 0);
```
svenruppert, line 28: extract in const, please

PR #788, opened 2023-08-24
Maven central

mvsoder:
Create scripts to release the artifacts and to generate the artifact bundle for Maven Central deployment


subprojects/testfx-core/testfx-core.gradle
```diff
@@ -16,7 +16,8 @@
  */
 apply plugin: 'java-library'
 
-ext.pomDescription = 'TestFX Core'
+ext.pomName = 'TestFX Core'
+ext.pomDescription = 'TestFX core library'
 ext.moduleName = 'org.testfx'
 ext.openjfxVersion = '12'
```
svenruppert, line 22: here is the version 12 in the other modules the version is 11 - why?

svenruppert (commented):
1) - see comment, please.
2) - add both of us to the developer sections, please.

subprojects/testfx-core/testfx-core.gradle
mvsoder: Probably just oversight. I will set this to 11 as the others are.

mvsoder:
Updates are complete

svenruppert (approved):
LGTM

